PR #1, opened 2024-04-22
registration/login + user management

SamuelVch98:


app.py
```diff
@@ -35,6 +36,65 @@ def index():  # put application's code here
 def private():
     return render_template("private.html")
 
+@app.route('/register')
```
anthonygego: If you plan to process each action in a specific route, I'd recommend you to separate related routes to a blueprint (see how authentication blueprint is implemented). This to avoid gathering all the code in a unique file.

Another way to spread the app code is to rely on *Class-based Views*. In this case, you can easily put each view is a separate file but you cannot use route decorators anymore.

app.py
```diff
@@ -35,6 +36,65 @@ def index():  # put application's code here
...
+def register():
+    return render_template('register.html')
+
+
+@app.route('/add_user', methods=['POST'])
+@login_required
+def add_user():
+    form = request.form
+    if form:
+        name = request.form['name']
```
anthonygego: SQLAlchemy automatically prevents SQL injection but that would be good to perform a basic sanity check on fields. For instance, check that the email format is correct, as the app will notify people per email. I guess names should also only allow a subset of valid characters.

db.py
```diff
@@ -0,0 +1,19 @@
+from sqlalchemy import create_engine, Column, Integer, String, ForeignKey, Boolean
```
anthonygego: You can rather use the `Flask-SQLAlchemy` package. This one allows to register the database session into the Flask globals (and therefore limit the amount of sessions to the amount of app threads instead of opening one for each data processing), but also to rely on the Flask configuration file for the database credentials.

templates/home.html
```diff
@@ -0,0 +1,50 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Home Page</title>
+    <!-- Bootstrap CSS -->
+    <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.0-alpha1/dist/css/bootstrap.min.css" rel="stylesheet">
```
anthonygego: You can copy and redistribute the minified Bootstrap package in a `/static` folder. This ensures service continuity in case the CDN is temporarily down, has changed, or simply maintainer has removed the packages (that happens).

app.py
```diff
@@ -35,6 +36,65 @@ def index():  # put application's code here
...
+def profile():
+    return render_template("profile.html")
+
+
+@app.route('/update_profile', methods=['POST'])
+@login_required
+def update_profile():
+    form = request.form
+    if form:
+        name = request.form['name']
```
anthonygego: Same remark here about sanitisation.

db.py
```diff
@@ -0,0 +1,19 @@
...
+
+    id = Column(Integer, primary_key=True)
+    name = Column(String, nullable=False)
+    first_name = Column(String, nullable=False)
+    email = Column(String, nullable=True)
+    admin = Column(Boolean, default=False)
+
+
+# Création de la base de données
+engine = create_engine('sqlite:///ictm-teaching.db', echo=True)
```
anthonygego: Using Flask-SQAlchemy will prevent from hardcoding the database URL.

templates/profile.html
```diff
@@ -0,0 +1,122 @@
...
+                </div>
+                <div class="tab-pane" id="preferences">
+                    <h2>Preferences for next year</h2>
+                    <hr>
+                </div>
+            </div>
+        </div>
+    </div>
+</div>
+<script src="https://cdn.jsdelivr.net/npm/@popperjs/core@2.11.6/dist/umd/popper.min.js" integrity="sha384-s9ftPkwp8Kt7sTv5yUpQnAwUxx/hdmZlpVhxAmRcxrI1vwNYmymgZb1t31cf3Zx+" crossorigin="anonymous"></script>
```
anthonygego: Include `popper.js` only if needed.

templates/profile.html
```diff
@@ -0,0 +1,122 @@
...
+<script src="https://cdn.jsdelivr.net/npm/bootstrap@5.3.0-alpha1/dist/js/bootstrap.bundle.min.js" integrity="sha384-Gcg0zY7Si34bQMQcj4RvDJB7RiJrmeiz26Ly+Qt8D4xjsHrh5zBoiyPz"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.6.0/jquery.min.js"></script>
```
anthonygego: It seems you import several versions of JQuery at different times in the same page. If you are actually using JQuery (it doesn't seem so right now), please import only one version and only once.

For the recall, Bootstrap 5 does not require JQuery anymore and embeds its own Javascript routines.

templates/profile.html
```diff
@@ -0,0 +1,122 @@
...
+<script src="https://cdn.jsdelivr.net/npm/bootstrap@5.3.0-alpha1/dist/js/bootstrap.bundle.min.js"></script>
```
anthonygego: That's probably not a good idea to use an alpha version of a library in such a project.

templates/register.html
```diff
@@ -0,0 +1,77 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Formulaire d'inscription</title>
+  <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.5.2/css/bootstrap.min.css">
```
anthonygego: It seems there is a Bootstrap version mismatch here.

templates/register.html
```diff
@@ -0,0 +1,77 @@
...
+  <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.0-alpha1/dist/css/bootstrap.min.css" rel="stylesheet">
+  <style>
+    body {
+      background-color: #f8f9fa;
```
anthonygego: Probably there is a Bootstrap class for doing what you want.

templates/register.html
```diff
@@ -0,0 +1,77 @@
...
+        <input type="text" class="form-control" id="firstname" name="first_name" placeholder="Entrez votre prénom">
+      </div>
+      <div class="form-group">
+        <label for="lastname">Nom</label>
+        <input type="text" class="form-control" id="name" name="name" placeholder="Entrez votre nom">
+      </div>
+      <input type="submit" value="Add">
+    </form>
+  </div>
+  <script src="https://stackpath.bootstrapcdn.com/bootstrap/4.5.2/js/bootstrap.min.js"></script>
```
anthonygego: Again check the relevance and versions of your imports.

templates/register.html
```diff
@@ -0,0 +1,77 @@
...
+        <a href="{{ url_for('auth.logout') }}" class="btn btn-outline-danger ms-2">Logout</a>
+      </div>
+    </div>
+  </div>
+</nav>
+  <div class="container">
+    <h2 class="text-center mb-4">New user</h2>
+    <form method="post" action="{{ url_for("add_user") }}">
+      <div class="form-group">
+        <label for="email">Adresse email</label>
```
anthonygego: French spotted here. Please stick to a unique language.

templates/layout.html
```diff
@@ -6,7 +6,8 @@
 </head>
 <body>
     {% if session.logged_in %}
-        <p>Hello {{ session.first_name }} {{ session.name }}. <a href="{{ url_for('auth.logout') }}">Log out</a></p>
+        <!--- <p>Hello {{ session.first_name }} {{ session.name }}. <a href="{{ url_for('auth.logout') }}">Log out</a></p> --->
+        {%  extends "home.html" %}
```
anthonygego: This is probably not what you meant to do. Template inheritance in Jinja works as follow :

layout.html :

    <html>
    <head>
        <script src="bootstrap.js"></script>
        <link href="bootstrap.css" rel="stylesheet" />
        {% block additionalpageheader %}{% endblock %}
        <title>{% block pagetitle %}{% endblock %} | ICTM Teaching</title>
    </head>
    <body>
        <div id="menu">
             
        </div>
        <div id="content">
            {% block pagecontent %}{%endblock %}
        </div>
    </body>
    </html>

home.html:

    {% extends "layout.html" %}
    {% block pagetitle %}Welcome{% endblock %}
    {% block pagecontent %}Hello world ! Please navigate this app trough the menu.{% endblock %}

You should not repeat the basic layout code in each template but only adds what is specific to the view. Block `additionalpageheader` can be defined in the child templates to add specific JS or CSS files only required for a specific view. The same thing can be done for the footer.

auth.py
```diff
@@ -33,6 +34,22 @@ def metadata():
...
+    db_session = Session()
+    try:
+        is_user = db_session.query(User).filter(User.email == user.email).first()
+        if is_user is None:
+            user = User(name=user.name, first_name=user.first_name, email=user.email)
+            db_session.add(user)
+            db_session.commit()
+            is_user = db_session.query(User).filter(User.email == user.email).first()
+        return is_user.id
+        return
```
anthonygego: Is this `return` reachable/useful ?

auth.py
```diff
@@ -33,6 +34,22 @@ def metadata():
...
 
+def check_or_create_user(user):
+    db_session = Session()
+    try:
+        is_user = db_session.query(User).filter(User.email == user.email).first()
+        if is_user is None:
+            user = User(name=user.name, first_name=user.first_name, email=user.email)
+            db_session.add(user)
+            db_session.commit()
+            is_user = db_session.query(User).filter(User.email == user.email).first()
```
anthonygego: You may be able to avoid duplicating the request by doing `db_session.refresh(user)`.

anthonygego (changes requested):
There is a major issue with your templating and dependencies. Other than that, these are suggestions to improve code maintainability once the codebase become larger.

Do not forget to update the `requirements.txt` file as you add Python package dependencies. You can hardcode versions using `==` there to be sure library updates don't break the code.

anthonygego:
@SamuelVch98 You can sign the CLA here : https://cla-assistant.io/UCL-INGI/ictm-teaching 
This is the FSF default template. That should be enough.

user.py
```diff
@@ -0,0 +1,90 @@
...
+@login_required
+def profile():
+    return render_template("profile.html")
+
+
+@user_bp.route('update_profile', methods=['POST'])
+@login_required
+def update_profile():
+    form = request.form
+    if form:
```
```suggestion
    if not form:
        return make_response("Problem with form request", 500)
```

and then de-indent code.

user.py
```diff
@@ -0,0 +1,90 @@
...
+    return render_template('register.html')
+
+
+def is_valid_email(email):
+    email_regex = r'^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$'
+    return re.match(email_regex, email) is not None
+
+
+def contains_valid_characters(name):
+    name_regex = r'^[a-zA-Z\-\' ]+$'
```
anthonygego: It seems you use this regex to check user names. Make sure accents are supported. There are specific identifier for alphabetical words `\w`.

user.py
```diff
@@ -0,0 +1,90 @@
...
+    return re.match(name_regex, name) is not None
+
+
+@user_bp.route('/add_user', methods=['POST'])
+@login_required
+def add_user():
+    form = request.form
+    if form:
```
anthonygego: Same here. `if not`, `return`, and then de-indent code.

templates/profile.html
```diff
@@ -0,0 +1,75 @@
...
+    <div class="container bootstrap snippet">
+        <div class="row">
+            <div class="col-sm-10">
+                <h1>{{ session.first_name }} {{ session.name }}</h1>
+            </div>
+        </div>
+        <div class="row">
+            <div class="col-sm-3">
+                <div class="text-center">
+                    <img src="http://ssl.gstatic.com/accounts/ui/avatar_2x.png" class="avatar img-circle img-thumbnail"
```
anthonygego: You can probably use fontawesome or something else here, rather than importing images whose links can change. Moreover I actually don't know what the license is but Google graphical elements are probably not meant to be used by third party projects.

templates/layout.html
```diff
@@ -2,14 +2,53 @@
...
+                    aria-controls="navbarNav" aria-expanded="false" aria-label="Toggle navigation">
+                <span class="navbar-toggler-icon"></span>
+            </button>
+            <div class="collapse navbar-collapse" id="navbarNav">
+                <ul class="navbar-nav">
+                    <li class="nav-item">
+                        <a class="nav-link" href="{{ url_for('user.profile') }}">My Profile</a>
+                    </li>
+                    <!---
+        <li class="nav-item dropdown">
```
anthonygego: Please avoid pushing commented code.

templates/register.html
```diff
@@ -0,0 +1,39 @@
+{% extends "layout.html" %}
+{% block pagetitle %}Register{% endblock %}
+{% block additionalpageheader %}{% endblock %}
```
anthonygego: There is no need to define a block in a child template if there is no content to put in there.

config.json
```diff
@@ -0,0 +1,40 @@
+{
```
anthonygego: Please do not push `config.json` but rather update `config.example.json` with example data and appropriate comments, or join a documentation.

app.py
```diff
@@ -1,34 +1,32 @@
...
-from functools import wraps
-from flask import Flask, render_template, redirect, session
 
-import auth
-from auth import auth_bp
 
 app = Flask(__name__)
-
 app.config.from_file("config.json", load=json.load)
+app.config['SQLALCHEMY_DATABASE_URI'] = 'sqlite:///ictm-teaching.db'
```
anthonygego: The goal of using a configuration file is actually to put every configuration field that may be deployment-specific **in** the file :grinning:

app.py
```diff
@@ -1,34 +1,32 @@
...
-    @wraps(f)
-    def decorated_function(*args, **kwargs):
-        if not session.get("logged_in", False):
-            return auth.login()
-        else:
-            return f(*args, **kwargs)
-    return decorated_function
-
+db.init_app(app)
+with app.app_context():
```
anthonygego: Watch out, it may be more careful to create a standalone script to create the database. I guess that in the most optimistic scenario, SQLAlchemy issues `IF NOT EXISTS` SQL commands but if your database schema changes it may create the new tables without updating the older, resulting in a broken database.

anthonygego (changes requested):
This PR will be small so we can squash the resulting diff into one unique commit, but you'd rather learn how to squash commits yourself for the future. This is useful to remove unnecessary things (or things that shouldn't have been committed) from the commit history, for instance the config file.

app.py
```diff
@@ -1,33 +1,33 @@
...
+from flask_sqlalchemy import SQLAlchemy
+import json
 
 app = Flask(__name__)
-
 app.config.from_file("config.json", load=json.load)
 
-# Core blueprints
-app.register_blueprint(auth_bp, url_prefix="/auth")
+with open('config.json') as config_file:
```
anthonygego: The line above already parses the configuration file and save it into the `app.config` dict. You only need to put `SQLALCHEMY_DATABASE_URI` into that file.

Reading your diff before submitting it could help you prevent such code smell.

SamuelVch98: Yes, it's my fault, I wanted to go too fast.

auth.py
```diff
@@ -43,12 +59,20 @@ def callback():
...
-        session["first_name"] = attrs["givenName"][0] if len(attrs["givenName"]) else ''
-        session["name"] = attrs["sn"][0]
-        session["email"] = attrs["email"][0]
+        session["uid"] = uid
+        session["first_name"] = first_name
+        session["name"] = name
+        session["email"] = email
 
+        user = User(name=name, first_name=first_name, email=email)
+        user_id = check_or_create_user(user)
```
anthonygego: Here you will always set the session name to the directory value at runtime. People that log into the app will always see their name as the updated directory one. However, these changes are not saved in database.

If Michel Dupont changes their name to Michelle Dupont, Michelle will see their name updated, while the administrators or other users will still see them as Michel, leading to a consistency issue.

I'd recommend, either:
- Always update user names from the directory value and trust identity based on returned email address and disallow users to change their names manually.
- Set the session name to the one stored in database. All users will see a consistent name and will have to update their name from the app if their name changes.

create_db.py
```diff
@@ -0,0 +1,11 @@
+from app import app, db
+
+
+def create_database():
```
anthonygego: I think we should create the admistrator account here. According to the chosen method for the issue above:
- Use `ArgumentParser` to process firstname, surname and email for the admin account
- Use email and blank (first-)names values that will be updated at first login

Please also provide a short notice in the README how to run the app for the first time.

requirements.txt
```diff
@@ -1,2 +1,3 @@
 flask
```
anthonygego: You can set the versions for all packages here, but not mandatory.

user.py
```diff
@@ -0,0 +1,89 @@
...
+    return render_template('register.html')
+
+
+def is_valid_email(email):
+    email_regex = r'^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$'
+    return re.match(email_regex, email) is not None
+
+
+def contains_valid_characters(name):
+    name_regex = r'^[\wÀ-ÖØ-öø-ÿ\'\- ]+$'
```
anthonygego: Doesn't `\w` matches Unicode characters by default ?

    >>> 'À'.isalpha()
    True

anthonygego (changes requested):

auth.py
AlexandreDoneux: By the directory value, do you mean the value sent by the SAML idp ?

requirements.txt
AlexandreDoneux: I think it would be a good thing. Try to use specific versions instead of minimum versions to be sure all the packages interact as intended. Incompatibilities could appear with new package versions.

templates/layout.html
```diff
@@ -2,14 +2,44 @@
 <html lang="en">
 <head>
     <meta charset="UTF-8">
-    <title>Title</title>
+    <title>{% block pagetitle %}{% endblock %} | ICTM Teaching</title>
+    <link rel="stylesheet" type="text/css" href="../static/bootstrap.min.css">
```
```suggestion
    <link rel="stylesheet" type="text/css" href="{{ url_for('static', filename='bootstrap.min.css') }}">
```

templates/layout.html
```diff
@@ -2,14 +2,44 @@
...
+    <script src="../static/bootstrap.min.js"></script>
```
```suggestion
    <script src="{{ url_for('static', filename='bootstrap.min.js') }}"></script>
```

anthonygego (approved):
I think we're good :-) 

I suggest to replace links in `layout.html` by `url_for` calls. Current URLs are actually not correct, the browser solves the issue itself.

I let @AlexandreDoneux have a last check and merge this one.

PR #5, opened 2024-06-25
Meeting 15 may copy

SamuelVch98:
- Updating the code to take account of the comments made by the first PR
- data.py : Add a 'data.py' file containing important data not stored in the DB
- templates/user_profile.html : merge tables to add preferences in a single dynamic table
    - See only the courses in your organisation in the preferences
- templates/course_info.html + db.py : delete the 'load_needed' field in the database and in the view 
- templates/courses.html : display the teachers for each course
- templates/layout.html : creation of drop-down menus for users and the course + creation of a confirmation modal for the change of year


templates/add_course.html
```diff
@@ -0,0 +1,110 @@
...
+    </style>
+    <script>
+        $(document).ready(function () {
+            $('#organization_code').change(function () {
+                var selectedOption = $(this).find('option:selected');
+                var selectedValue = selectedOption.val();
+                var selectedText = selectedOption.text();
+
+                if (selectedValue) {
+                    var tagHtml = '<div class="badge bg-primary d-inline-block" data-id="' + selectedValue + '">' +
```
anthonygego: jQuery has methods to construct HTML elements without coding into strings...

    jQuery('<div>', {
        id: '...',
        class: '...',
    }).appendTo('#...');

app.py
```diff
@@ -1,21 +1,46 @@
...
+
+def get_configurations():
+    if session.get('logged_in'):
+        return db.session.query(Configuration).all()
+    else:
+        return []
+
+
+def get_organization():
+    if session.get('logged_in'):
```
anthonygego: Is there any case where you need to use the organisations code without being logged ? Otherwise this shouldn't be necessary.

auth.py
```diff
@@ -74,6 +73,10 @@ def callback():
         session["user_id"] = updated_user.id
         session["first_name"] = updated_user.first_name
         session["name"] = updated_user.name
+        session["is_admin"] = updated_user.admin
+
+        config = db.session.query(Configuration).filter_by(is_current_year=True).first()
+        session["current_year"] = config.year
```
anthonygego: Is there no way to compute the current year dynamically based on the actual year we are in ? This would avoid saving current year status.
I d not mean we absolutely don't need to store theyear in session: that may probably be useful to go into the past for administrators. However, some arguments may be more relevant in the app routes than in session.

config.py
```diff
@@ -0,0 +1,48 @@
...
+@config_bp.route('/next_year', methods=['POST'])
+@login_required
+def next_year():
+    last_year = Configuration.query.order_by(Configuration.year.desc()).first()
+
+    if last_year is None:
+        return make_response("There are no years in the database yet", 500)
+
+    current_year = last_year.year
+    start_year, end_year = map(int, current_year.split('-'))
```
anthonygego: It seems a bit cumbersome. You can simply store the year as an int and consider a bijection between year X and academic year (X-1)-X or X-(X+1) when showing data to the user.

config.py
```diff
@@ -0,0 +1,48 @@
...
+config_bp = Blueprint('config', __name__)
+
+
+@config_bp.route('/next_year', methods=['POST'])
+@login_required
+def next_year():
+    last_year = Configuration.query.order_by(Configuration.year.desc()).first()
+
+    if last_year is None:
+        return make_response("There are no years in the database yet", 500)
```
anthonygego: When is this supposed to happen ? Can't we create the first one, if needed, when the database is inistialized ?

config.py
```diff
@@ -0,0 +1,48 @@
...
+    except Exception as e:
+        db.session.rollback()
+        raise e
+
+    return redirect(url_for("index"))
+
+
+@config_bp.route('/update_current_year', methods=['POST'])
+@login_required
+def update_current_year():
```
anthonygego: Same comment as before here. Maybe it's simpler to put the year we want to grab data into the route. The advantage of proceeding like this is that you directly get the year from the method argument and cannot be mistaken. 

You also avoid very weird issues. One such issue could be that you open two pages of the app into two different tabs. You apply the change in one tab, and the session gets updated. The older tab will not reflect the year change and you may end up modifying the wrong data without getting an error.

course.py
```diff
@@ -0,0 +1,223 @@
+from decorators import login_required
+from db import db, User, Course, Teacher, Organization
+from data import LANGUAGES, QUADRI
+from flask import Blueprint, render_template, flash, url_for, request, make_response, redirect, \
+    Flask, jsonify, session
+import json, re
+
+course_bp = Blueprint('course', __name__)
+
+app = Flask(__name__)
```
anthonygego: Why do you reload the app here ?

course_preference.py
```diff
@@ -0,0 +1,78 @@
+from decorators import login_required
+from db import db, Researcher, PreferenceAssignment
+from flask import Blueprint, render_template, flash, url_for, request, make_response, redirect, \
+    Flask, jsonify, session
+import json
+
+course_preference_bp = Blueprint('course_preference', __name__)
+
+app = Flask(__name__)
```
anthonygego: Why do you reload the app here ?

course.py
```diff
@@ -0,0 +1,223 @@
...
+
+def validate_course_title(title):
+    title_regex = r'^[\w\s,:\']+$'
+    return re.match(title_regex, title) is not None
+
+
+@course_bp.route('/form_course')
+@login_required
+def form_course():
+    return render_template('add_course.html', quadri=QUADRI, language=LANGUAGES)
```
anthonygego: You push configuration and organisations code into the template globals, maybe you want to do the same things for both. Both methods are OK though, but it would be clearer to choose one ans stick to it.

config.py
```diff
@@ -0,0 +1,48 @@
+from decorators import login_required
+from db import db, Configuration
+from flask import Blueprint, render_template, flash, current_app, url_for, request, make_response, redirect, session, \
+    Flask
+
+config_bp = Blueprint('config', __name__)
+
+
+@config_bp.route('/next_year', methods=['POST'])
+@login_required
```
anthonygego, line 10: Who is supposed to add a new year ? It seems there is no permission check here.

course.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+@course_bp.route('/add_course', methods=['POST'])
+@login_required
```
anthonygego, line 69: It seems there is no permission checks in the app right now, do not forget to add them at the right locations.

course.py
```diff
@@ -0,0 +1,223 @@
...
+
+    code = request.form['code']
+    title = request.form['title']
+
+    if not validate_course_code(code):
+        return make_response("Invalid course code", 400)
+    if not validate_course_title(title):
+        return make_response("Invalid course title", 400)
+
+    quadri = request.form['quadri']
```
anthonygego, line 84: Performing validation on `code` and `title` is right, what about the other fields ?

course.py
```diff
@@ -0,0 +1,223 @@
...
+    year = request.form['year']
+    language = request.form['language']
+    organization_ids = request.form.getlist('organization_code[]')
+
+    try:
+        last_course = db.session.query(Course).order_by(Course.id.desc()).first()
+        new_id = 1 if last_course is None else last_course.id + 1
```
anthonygego: What about auto-incrementing ?

course.py
```diff
@@ -0,0 +1,223 @@
...
+def courses():
+    current_year = session["current_year"]
+    courses = db.session.query(Course).filter_by(year=current_year).all()
+    return render_template('courses.html', courses=courses)
+
+
+@course_bp.route('/search_teachers')
+def search_teachers():
+    search_term = request.args.get('q', '')
+    teachers = db.session.query(User).filter(User.is_teacher == 1, User.name.ilike(f'%{search_term}%')).all()
```
anthonygego: Make sure no injection are possible here.

course.py
```diff
@@ -0,0 +1,223 @@
...
+@login_required
+def course_info(course_id):
+    current_year = session["current_year"]
+    course = db.session.query(Course).filter(Course.id == course_id, Course.year == current_year).first()
+    if not course:
+        return make_response("Course not found", 404)
+
+    all_years = db.session.query(Course).filter_by(id=course.id).distinct(Course.year).order_by(
+        Course.year.desc()).all()
+    return render_template('course_info.html', course=course, all_years=all_years, quadri=QUADRI,
```
anthonygego: Same comments for globals and templates

course.py
```diff
@@ -0,0 +1,223 @@
...
+    nbr_students = request.form['nbr_students']
+    nbr_teaching_assistants = request.form['nbr_teaching_assistants']
+    nbr_monitor_students = request.form['nbr_monitor_students']
+
+    course = db.session.query(Course).filter(Course.id == course_id, Course.year == year).first()
+    if not course:
+        return make_response("Course not found", 404)
+
+    if assigned_teachers:
+        try:
```
anthonygego: Well maybe it's simpler in terms of requests to remove every one and add them back, even if there is an overlap in the data.

course.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+@course_bp.route('/duplicate_course')
+@login_required
+def duplicate_course():
+    course_id = request.args.get('course_id')
+    course_year = request.args.get('year')
+    course = db.session.query(Course).filter(Course.id == course_id, Course.year == course_year).first()
+
+    return render_template('duplicate_course.html', course=course, quadri=QUADRI, language=LANGUAGES)
```
anthonygego: Same comment for globals

course.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+@course_bp.route('/add_duplicate_course', methods=['POST'])
+@login_required
+def add_duplicate_course():
+    form = request.form
```
anthonygego, line 215: I've already seen this exact same block of 10+ lines of code, maybe you can write a common sanitizing method.

course_preference.py
```diff
@@ -0,0 +1,78 @@
...
+app.config.from_file("config.json", load=json.load)
+
+
+def delete_old_preferences(researcher_id, course_ids):
+    current_year = session["current_year"]
```
anthonygego: It would be better not to make your utility functions depend on the context.

course_preference.py
```diff
@@ -0,0 +1,78 @@
...
+    preferences = data['preferences']
+    if preferences is None:
+        return make_response("No preferences received", 500)
+
+    user_id = session["user_id"]
+    researcher = Researcher.query.filter(Researcher.user_id == user_id).first()
+    if researcher is None:
+        return make_response("User is not a researcher", 500)
+
+    new_course_ids = {preference.split("-", 1)[0] for preference in preferences}
```
anthonygego: If your POST request payload is a JSON encoded dict, maybe that would be nice to take advantage of tis complex structure and avoid encoding tuple/lists with dashes that you need to postprocess.

create_db.py
```diff
@@ -21,6 +24,34 @@ def add_first_admin(name, first_name, email):
...
+
+    if existing_year is None:
+        config = Configuration(year=current_year)
+        db.session.add(config)
+        db.session.commit()
+
+
+# Create organizations
+def create_organizations():
+    organizations = ["SST", "ICTM", "ELEN", "EPL", "INMA", "SSH", "IMMC", "IMAQ", "INGI", "SSS", "IONS", "ELI",
```
anthonygego: Maybe we should put them in the configuration file rather than hardcoding them.

create_db.py
```diff
@@ -21,6 +24,34 @@ def add_first_admin(name, first_name, email):
         print("Admin account already exists.")
 
 
+def get_current_academic_year():
+    current_year = datetime.now().year
+    next_year = current_year + 1
+    academic_year = f"{current_year}-{next_year}"
```
anthonygego: So here you should simply keep `current_year` or `next_year`.

data.py
```diff
@@ -0,0 +1,6 @@
+LANGUAGES = {
```
anthonygego: Either this file should be named `constants` or `enums`, either it should go the config file. @AlexandreDoneux what do you think ? The first option seems the right one to me, as this will probably never change.

db.py
```diff
@@ -1,13 +1,150 @@
...
+
 class User(db.Model):
     __tablename__ = 'user'
     id = db.Column(db.Integer, primary_key=True)
     name = db.Column(db.String, nullable=True)
     first_name = db.Column(db.String, nullable=True)
-    email = db.Column(db.String, nullable=True)
+    email = db.Column(db.String, nullable=True, unique=True)
     admin = db.Column(db.Boolean, default=False)
+    is_teacher = db.Column(db.Boolean, default=False)
```
anthonygego, line 17: If you have models for Teacher and Researcher you should better only count on the relationship expressed in the latter models and do not keep this here. Relational databases are made to keep track of consistency.

db.py
```diff
@@ -1,13 +1,150 @@
...
+    @staticmethod
+    def before_update(mapper, connection, target):
+        if not target.active:
+            if target.supervisees:
+                raise ValueError("Cannot deactivate a supervisor who has supervisees.")
+            if target.user_teacher:
+                raise ValueError("Cannot deactivate a teacher assigned to a course.")
+
+
+db.event.listen(User, 'before_update', User.before_update)
```
anthonygego: Is there no way to express this natively via SQLAlchemy ?

templates/add_course.html
```diff
@@ -0,0 +1,110 @@
...
+            color: black;
+            text-decoration: none;
+        }
+
+        .badge a:hover {
+            color: red;
+        }
+    </style>
+    <script>
+        $(document).ready(function () {
```
anthonygego: You do not *have to* put this in the header if it needs to run only when the document is ready. You can put this wherever you want, or at the bottom of the page to accelerate the page rendering. However, the content is very small and that won't have any perceptible effect.

templates/course_info.html
```diff
@@ -0,0 +1,238 @@
...
+            color: red;
+        }
+    </style>
+    <script>
+        $(document).ready(function () {
+            function formatTeacher(teacher) {
+                if (teacher.loading) {
+                    return teacher.text;
+                }
+                var $container = $("<div></div>").text(teacher.text);
```
anthonygego: Beware of the `$` symbol, i can be confusing when using JQuery as a framework.

templates/course_info.html
```diff
@@ -0,0 +1,238 @@
...
+                return $container;
+            }
+
+            $('[id^=assigned_teachers]').each(function () {
```
anthonygego: You can use classes and the dot selector if you have multiple elements to match.

templates/course_info.html
```diff
@@ -0,0 +1,238 @@
...
+                    },
+                    templateResult: formatTeacher,
+                    templateSelection: formatTeacher,
+                    tags: false
+                });
+            });
+        });
+    </script>
+    <script>
+        $(document).ready(function () {
```
anthonygego: You can merge both the `$(document).ready()` blocks

templates/course_info.html
```diff
@@ -0,0 +1,238 @@
...
+            $('#organization_code').change(function () {
+                var selectedOption = $(this).find('option:selected');
+                var selectedValue = selectedOption.val();
+                var selectedText = selectedOption.text();
+
+                if (selectedValue) {
+                    var tagHtml = '<div class="badge bg-primary" data-id="' + selectedValue + '">' +
```
anthonygego: Same remark: you can use JQuery to build HTML elements from a dict. BTW, this piece of code looks similar to the one in `add_course.html`. You can make a JS utility script when common functions.

templates/courses.html
```diff
@@ -0,0 +1,35 @@
...
+            <thead>
+            <tr>
+                <th>Title</th>
+                <th>Teachers</th>
+                <th>Organizations</th>
+            </tr>
+            </thead>
+            <tbody>
+            {% for course in courses %}
+                <tr onclick="window.location='{{ url_for("course.course_info", course_id=course.id) }}';"
```
anthonygego: Use a `<a>`-based Bootstrap listgroup here, instead on binding JS code to act as a link.

templates/duplicate_course.html
```diff
@@ -0,0 +1,213 @@
...
+            color: black;
+            text-decoration: none;
+        }
+
+        .badge a:hover {
+            color: red;
+        }
+    </style>
+    <script>
+        $(document).ready(function () {
```
anthonygego: This whole 50+ lines script looks similar to the one in `course_info.html`. Please refactor.

templates/duplicate_course.html
```diff
@@ -0,0 +1,213 @@
...
+            $(document).on('click', '.remove-tag', function (e) {
+                e.preventDefault();
+                $(this).parent('.badge').remove();
+            });
+        });
+    </script>
+{% endblock %}
+
+{% block pagecontent %}
+    <div class="row">
```
anthonygego: This whole template looks a lot like `course_info.html`. Maybe you can refactor both in order to re-use common blocks.

templates/index.html
```diff
@@ -1,7 +1,7 @@
-{%  extends "layout.html" %}
```
anthonygego, line 1: Do you still need this template ?

templates/layout.html
```diff
@@ -4,32 +4,140 @@
...
+    <script src="{{ url_for('static', filename='select2.min.js') }}"></script>
+
+    <style>
+        .form-select {
+            width: auto !important;
+        }
+    </style>
+    <script>
+        $(document).ready(function () {
+            $('#newYear').click(function () {
```
anthonygego: You can perform this bindng directly in the HTML code.

templates/layout.html
```diff
@@ -4,32 +4,140 @@
...
+                $('#confirmationModal').modal('show');
+            });
+
+            $('#confirmChange').click(function () {
```
anthonygego: You can also put a POST form into the modal code and use a submit button, that would greatly simplify the code.

templates/private.html
```diff
@@ -1,5 +1,5 @@
-{%  extends "layout.html" %}
```
anthonygego, line 1: Do you need this template anymore ?

templates/user_profile.html
```diff
@@ -0,0 +1,377 @@
...
+                $('#staticPreferencesTable').show();
+                $('#addPreferencesButton').show();
+                $('#savePreferencesButton').hide();
+                $('#cancelEditButton').hide();
+                $('#addRowButton').hide();
+            });
+
+            $('#addRowButton').click(function () {
+                const newRow = `
+                    <tr>
```
anthonygego: I strongly recommend you to transform this piece of code as a hidden HTML element and use it as a template. You select the template using JQuery, copy it and set the data before inserting it at the right location.

templates/user_profile.html
```diff
@@ -0,0 +1,377 @@
...
+                const preferences = [];
+                $('#editablePreferencesTable tr').each(function () {
+                    const courseCode = $(this).find('select.course-select').val();
+                    if (courseCode) {
+                        preferences.push(courseCode);
+                    }
+                });
+
+                const data = {
+                    preferences: preferences
```
anthonygego: Do you actually need to encapsulate your preferences into a one-key dict ?

templates/users.html
```diff
@@ -0,0 +1,100 @@
...
+                {% endif %}
+                {% if user_type == 'archived' %}
+                    <th>Type</th>
+                {% endif %}
+                <th></th>
+            </tr>
+            </thead>
+            <tbody>
+            {% for user in users %}
+                <tr onclick="window.location='{{ url_for("user.user_profile", user_id=user.id) }}';"
```
anthonygego: Same remark as above, use a Bootstrap listgroup with `<a>` tags instead.

user.py
```diff
@@ -1,15 +1,31 @@
...
+    Flask
+import re, json
 
 user_bp = Blueprint('user', __name__)
 
+app = Flask(__name__)
+app.config.from_file("config.json", load=json.load)
+
+
+def get_default_max_load(researcher_type):
```
anthonygego: Maybe an enum would be more resilient as runtime as it won't even load if the enum does not exist.

user.py
```diff
@@ -1,15 +1,31 @@
...
-import re
+from db import db, User, Researcher, Course, PreferenceAssignment
+from data import RESEARCHERS_TYPE
+from flask import Blueprint, render_template, flash, url_for, request, make_response, redirect, session, \
+    Flask
+import re, json
 
 user_bp = Blueprint('user', __name__)
 
+app = Flask(__name__)
```
anthonygego: Why do you need to reload the app ?

user.py
```diff
@@ -32,33 +48,84 @@ def add_user():
...
+    researcher_type = request.form['researcher_type'] if is_researcher else None
+
     try:
-        new_user = User(name=name, first_name=first_name, email=email)
+        new_user = User(name=name, first_name=first_name, email=email, is_teacher=is_teacher,
+                        is_researcher=is_researcher, supervisor_id=supervisor_id,
+                        organization_id=organization_code)
         db.session.add(new_user)
         db.session.commit()
+        if is_researcher:
```
anthonygego, line 58: I suspect you need to keep such a field in the `User` model to do this. Actually, creating a teacher or researcher profile can be done in two steps if needed, or just be processed at runtime and not stored in database.

user.py
```diff
@@ -32,33 +48,84 @@ def add_user():
...
+    return render_template('users.html', users=all_users, user_type=user_type)
 
 
-@user_bp.route('update_profile', methods=['POST'])
+@user_bp.route('/profile/<int:user_id>')
 @login_required
-def update_profile():
+def user_profile(user_id):
+    all_users = db.session.query(User).filter(User.admin == 0, User.is_teacher == 1, User.active == 1).all()
+    my_user = db.session.query(User).filter_by(id=user_id).first()
```
anthonygego: Maybe `my_user` is a bad name as it contains the requested user, while `my_profile` is really session-specific.

user.py
```diff
@@ -74,16 +141,70 @@ def update_profile():
...
-            session["first_name"] = first_name
-            session["name"] = name
-        except Exception as e:
-            db.session.rollback()
-        return redirect(url_for("user.profile"))
-    else:
-        flash('User not found', 'danger')
+    user_id = request.form['user_id']
+    organization_code = None if request.form['organization_code'] == 'None' else request.form['organization_code']
+    is_teacher = 1 if 'is_teacher' in request.form else 0
```
anthonygego: Aren't booleans supported here ?

anthonygego (changes requested):
This is a very large PR that could have been splitted in multiple ones. It took more  than 2 hours of reading and is not comprehensive.

It seems that permission are not implemented yet, every logged user being able to perform every action. That should be addressed in the next PR.

It won't be possible to have a nice commit log for this one but we expect the commit log to be nice and clean for the future steps and, most of all, the PRs to be smaller. This one will probably take time to be mergeable due to its length because each iteration may take as many time.

I hope @AlexandreDoneux have some time to perform his own pass on the code.

data.py
AlexandreDoneux: I agree. I told him those values will most likely not change so putting them in the config file is not necessary. I also think another name for the file might be good.

course.py
SamuelVch98: The other fields are values from drop down, so I assumed they didn't need to be checked

course.py
SamuelVch98: The problem was that we had decided that to identify the same course it had to have the same id. If I set an auto-incrementing id, it's not possible to create a course with the same id (in the case of course dupplication). Perhaps a solution is possible but I haven't found one.

config.py
SamuelVch98: this will be done in the next PR as indicated in your first comment

course.py
SamuelVch98: this will be done in the next PR as indicated in your first comment

templates/user_profile.html
SamuelVch98: I used a dictionary because for each preference I have to save the year of the course and its id. Is there any other way?

user.py
SamuelVch98: I spoke to Alexandre about how to implement this problem and he advised me that when I create a user and check its role, I create a tuple in the db in this way: 

If it's a teacher: I create a tuple that only contains the user id linked to the time that the user is assigned to a lesson. 
If it's a researcher: I create the researcher directly because I don't need any specific conditions

In addition to this, I've added an 'is_active' field in the researcher and teacher table, which means I don't have to delete the row in the db every time I check/uncheck the teacher/researcher box.

Does this solution seem right to you or is there a simpler way of solving this problem?

course.py
AlexandreDoneux: When looking at your code I think you just grab the last course id added and increment it. I don't see where, in the case of a duplicate, you check for the existing course and use the same id. 

What you need to do is go fetch the id of the course you are duplicating and change only the year (and possibly code, title, quadri and language). You are just fetching the last id, no matter the course, and increment it. That is something you need to do only when adding a non duplicate course.

Maybe SQLAlchemy allows you to define a column as "incremented when the value is not given". With this you would only need to fetch the id of the course if it already exists. If it doesn't exist the id would be automatically incremented. Look into it.

static/scripts/util.js
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// Function to handle the change event of the yearSelect dropdown
+function handleYearSelectChange(params) {
+    $('#yearSelect').change(function () {
+        params.selected_year = $(this).val();
+        let updatedUrl = "/config/update_current_year";
+
+        console.log("updateCurrentYearUrl:", updatedUrl);
+        console.log('Params:', params);
```
AlexandreDoneux: Why the different console.log in this file ? It would be better to not have them in the end product. And for the errors there are better ways to display errors but focus on the rest for the time being.

user.py
AlexandreDoneux: Samuel, I don't think I understand your explanation. What do you mean by tuple ? 

Essentially, a teacher/researcher is created in DB for every course the user is a teacher/researcher for, right ? So you don't need to create a Teacher or Researcher instance in the DB right away. That can be done when the user is assigned to a course (as teacher or researcher). 
Storing "is_researcher" and "is_teacher" in the DB allows us to give access to certain features/pages in the app. And that can later be changed to roles. Hadn't we discussed those flags in a previous meeting ?

SamuelVch98: I misunderstood Anthony's message and got everything mixed up in my code, so you can ignore my message.

course.py
SamuelVch98: My problem was specific to SQLite. I couldn't create a compound key that contained one of the PKs that is auto incremented. The problem was solved when I used Postgres

config.py
```diff
@@ -0,0 +1,45 @@
...
+    current_year = int(request.form.get('selected_year'))
+    endpoint = request.form.get('current_endpoint')
+
+    if endpoint == 'user.user_profile':
+        user_id = request.form.get('user_id')
+        url = url_for(endpoint, user_id=user_id, current_year=current_year)
+    else :
+        url = url_for(endpoint, current_year=current_year)
+
+    return url
```
anthonygego: Shouldn't this be a `redirect` as well ? I'm wondering if you couldn't simply handle this in JS.

If you keep this in Python, there is probably no need for the `if` statement, you can pass the `user_id` argument whatever the endpoint is except if you use this name for anything else in other routes.

course.py
```diff
@@ -0,0 +1,262 @@
...
+    pattern = r'^[a-zA-Z\s]+$'
+    return re.match(pattern, string) is not None
+
+
+def validate_number_pattern(number):
+    pattern = r'^(0|[1-9][0-9]*)$'
+    return re.match(pattern, number) is not None
+
+
+@course_bp.route('/form_course')
```
anthonygego: Maybe we keep keep the same route `add_course` and simply return the `add_course.html` template if the request method is a GET.

course.py
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    for field, validator in mandatory_fields.items():
+        value = form.get(field)
+        if not value:
+            return f"Missing {field}", 400
+        if not validator(value):
+            return f"Invalid {field}", 400
+
+    if extra_fields_needed:
```
anthonygego: You can simply add the extra fields dict to the mandatory one in this if statement to avoid repeating the validating code. This if statement would just go up and only contain one dictionary merge statement.

course.py
```diff
@@ -0,0 +1,262 @@
...
+@course_bp.route('/add_course', methods=['POST'])
+@login_required
+def add_course():
+    form = request.form
+    if not form:
+        return make_response("Problem with form request", 500)
+
+    error = validate_form_data(form)
+    if error:
+        return make_response(error[0], error[1])
```
```suggestion
        return make_response(*error)
```
The `*` operator explodes the tuple and allows to use the tuple values as positional arguments.
The `**` operator does the same thing for dictionaries and named arguments.

course.py
```diff
@@ -0,0 +1,262 @@
...
+        return redirect(url_for("course.form_course"))
+    except Exception as e:
+        db.session.rollback()
+        raise e
+
+
+@course_bp.route('/courses/<int:current_year>')
+@login_required
+def courses(current_year=None):
+    current_year = request.args.get('current_year') if request.args.get('current_year') else current_year
```
anthonygego: It seems very weird to have two ways of specifying the year to show. This means you can have URLs like `/courses/2023?current_year=2024` and that `2024` takes priority over `2023` if present. Why don't you simply use the route-embedded argument ?

course.py
```diff
@@ -0,0 +1,262 @@
...
+@course_bp.route('/update_course_info', methods=['POST'])
+@login_required
+def update_course_info():
+    form = request.form
+    if not form:
+        return make_response("Problem with form request", 500)
+
+    error = validate_form_data(form)
+    if error:
+        return make_response(error[0], error[1])
```
```suggestion
        return make_response(*error)
```

course.py
```diff
@@ -0,0 +1,262 @@
...
+                return f"Invalid {field}", 400
+
+def assign_teachers_to_course(course_id, course_year, assigned_teachers):
+    try:
+        for teacher_id in assigned_teachers:
+            existing_teacher = db.session.query(Teacher).filter(Teacher.course_id == course_id,
+                                                                Teacher.user_id == teacher_id,
+                                                                Teacher.course_year == course_year).first()
+
+            if existing_teacher is None:
```
anthonygego: If you have removed the existing entries before there should be no need to test if they exists before insertion.

course.py
```diff
@@ -0,0 +1,262 @@
...
+@course_bp.route('/add_duplicate_course', methods=['POST'])
+@login_required
+def add_duplicate_course():
+    form = request.form
+    if not form:
+        return make_response("Problem with form request", 500)
+
+    error = validate_form_data(form, extra_fields_needed=True)
+    if error:
+        return make_response(error[0], error[1])
```
```suggestion
        return make_response(*error)
```

create_db.py
```diff
@@ -21,6 +25,24 @@ def add_first_admin(name, first_name, email):
...
+    if existing_year is None:
+        config = Configuration(year=current_year)
+        db.session.add(config)
+        db.session.commit()
+
+
+# Create organizations
+def create_organizations():
+    if db.session.query(Organization).count() == 0:
+        db.session.add_all([Organization(name=org) for org in ORGANIZATIONS])
```
anthonygego, line 43: It seems a bit cumbersome to put into database entries that are harcoded in a enumeration. I suppose this enum is not used elsewhere in the code, in which case there could be a consistency issue. Maybe you can simply put the default listing here.

This would ensure that the enum is never actually used in the code in the future.

db.py
anthonygego: Is there any good reason for this ?

static/scripts/util.js
```diff
@@ -0,0 +1,68 @@
...
+        success: successCallback,
+        error: errorCallback
+    });
+}
+
+// Function to handle the change event of the yearSelect dropdown
+function handleYearSelectChange(params) {
+    $('#yearSelect').change(function () {
+        params.selected_year = $(this).val();
+        let updatedUrl = "/config/update_current_year";
```
anthonygego: So maybe you can handle the whole redirection thing here without asking the server for the url. You can for instance generate the current endpoint URL and replace the current year dynamically before changing the window location.

templates/layout.html
```diff
@@ -4,32 +4,105 @@
...
+                                                data-bs-target="#confirmationModal">Add New Year
+                                        </button>
+                                    </li>
+                                </ul>
+                            </li>
+                        {% endif %}
                         {% block additionalnav %}{% endblock %}
                     </ul>
                     <div class="navbar-text ms-auto">
+                        &nbsp;&nbsp;&nbsp;&nbsp;
```
anthonygego: Bootstrap provides several CSS classes to play with margins and paddings. That may help to avoid putting non-breakable spaces.

templates/users.html
```diff
@@ -0,0 +1,108 @@
...
+                {% if user_type == 'archived' %}
+                    <th>Type</th>
+                {% endif %}
+                <th></th>
+            </tr>
+            </thead>
+            <tbody>
+            {% for user in users %}
+                <tr>
+                    <td><a href="{{ url_for("user.user_profile", user_id=user.id, current_year=dynamic_year) }}">
```
anthonygego, line 55: Where is `dynamic_year` populated from ? I don't find its definition.

util.py
```diff
@@ -0,0 +1,10 @@
+import datetime
+
+
+def get_current_year():
+    current_date = datetime.date.today()
+    current_year = current_date.year
+    if current_date.month >= 9:  # Assuming academic year starts in September
```
anthonygego, line 7: This must be discussed with @luciledierckx but it's ok for me. Maybe we want the current year to start at an arbitrary moment.

user.py
```diff
@@ -1,15 +1,19 @@
...
+import re, json
 
 user_bp = Blueprint('user', __name__)
 
 
 @user_bp.route('/register')
 @login_required
 def register():
-    return render_template('register.html')
+    supervisors = db.session.query(User).filter(User.admin == False, User.is_teacher == True, User.active == True).all()
```
anthonygego: Is there any reason to filter out admins from the potential supervisors ?

user.py
anthonygego: For me there is still no need to keep `is_teacher` or `is_researcher` fields in the `User` model. You know a user is a teacher or researcher (or...both) if a `Teacher` or `Researcher` exists in database. SQLAlchemy relationships should handle this. Otherwise you need to keep the cosnsitency manually and check that the field in `User` table is correct when a profile is deleted.

This is not blocking. We can go forward with this right now.

anthonygego (changes requested):

anthonygego:
Those minor comments should be fixed quickly. Please fix them as soon as possible so that we can merge and be able to read the next PRs diffs correctly. Maybe you will need to rebase them once this one is merged.

config.py
SamuelVch98: I tried to use redirect but it didn't work. It didn't take the redirect into account, so I tried that way and it worked.

templates/users.html
SamuelVch98: dynamic_year is defined in the @app.context_processor

util.py
SamuelVch98: I asked the question

user.py
SamuelVch98: No particular reason, I didn't take into account the case where a teacher is an admin.

static/scripts/util.js
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// Function to handle the change event of the yearSelect dropdown
+function handleYearSelectChange(params) {
+    $('#yearSelect').change(function () {
+        const selectedYear = $(this).val();
+        let currentEndpoint = params.current_endpoint;
+        let parts = currentEndpoint.split('/');
+
+        parts[parts.length - 1] = selectedYear; // Update the year in the endpoint
```
anthonygego: I really think we should rather generate the `currentEndpoint` value from a call to `url_for` with a placeholder value for the year. This way if we change the route arguments position, this will be less prone to regressions.

templates/courses.html
```diff
@@ -0,0 +1,69 @@
...
+                    <select class="form-select form-select-sm d-inline" id="yearSelect" name="yearSelect"
+                            {% if session.is_admin == False %} disabled {% endif %}>
+                        {% for config in configurations %}
+                            <option value="{{ config.year }}" {% if config.year == current_year %}
+                                    selected {% endif %}>
+                                {{ config.year }}-{{ config.year + 1 }}
+                            </option>
+                        {% endfor %}
+                    </select>
+                    <input type="hidden" name="current_endpoint" id="current_endpoint" value="{{ url_for('course.courses', current_year=dynamic_year) }}">
```
anthonygego: So here you should rather set a placeholder instead of `dynamic_year`

templates/user_profile.html
```diff
@@ -0,0 +1,404 @@
...
+                                                {% if session.is_admin == False %} disabled {% endif %}>
+                                            {% for config in configurations %}
+                                                <option value="{{ config.year }}" {% if config.year == current_year %}
+                                                        selected {% endif %}>
+                                                    {{ config.year }}-{{ config.year + 1 }}
+                                                </option>
+                                            {% endfor %}
+                                        </select>
+                                        <input type="hidden" name="current_endpoint" id="current_endpoint"
+                                               value="{{ url_for('user.user_profile', user_id=requested_user.id, current_year=dynamic_year) }}">
```
anthonygego: Probably same thing here.

anthonygego (changes requested):

db.py
```diff
@@ -1,13 +1,151 @@
...
+                              secondary='course_organization',
+                              back_populates='organizations',
+                              primaryjoin="Organization.id == CourseOrganization.organization_id",
+                              secondaryjoin="and_(Course.id == CourseOrganization.course_id, Course.year == CourseOrganization.course_year)")
+
+
+class CourseOrganization(db.Model):
+    __tablename__ = 'course_organization'
+
+    course_id = db.Column(db.Integer, db.ForeignKey('course.id'), primary_key=True)
```
```suggestion
    course_id = db.Column(db.Integer, primary_key=True)
```

The ForeignKeyConstraint is specified below

db.py
```diff
@@ -1,13 +1,151 @@
...
+    course_year = db.Column(db.Integer, db.ForeignKey('course.year'), primary_key=True)
```
```suggestion
    course_year = db.Column(db.Integer, primary_key=True)
```

The ForeignKeyConstraint is specified below

course.py
```diff
@@ -0,0 +1,250 @@
...
+            return make_response("Course already exists", 500)
+
+        new_course = Course(year=year, code=code, title=title, quadri=quadri, language=language)
+        # Fetch organizations and add them to the course
+        organizations = db.session.query(Organization).filter(Organization.id.in_(organization_ids)).all()
+        new_course.organizations.extend(organizations)
+
+        db.session.add(new_course)
+        db.session.commit()
+        return redirect(url_for("course.form_course"))
```
```suggestion
        return redirect(url_for("course.courses", current_year=year))
```

course.py
```diff
@@ -0,0 +1,250 @@
...
+    except Exception as e:
+        db.session.rollback()
+        raise e
+
+
+@course_bp.route('/add_course', methods=['POST', 'GET'])
+@login_required
+def add_course():
+    if request.method == 'GET':
+        return redirect(url_for('course.form_course'))
```
```suggestion
        return render_template('add_course.html')
```

static/scripts/util.js
```diff
@@ -0,0 +1,61 @@
...
+        type: 'POST',
+        url: url,
+        data: data,
+        success: successCallback,
+        error: errorCallback
+    });
+}
+
+// Function to handle the change event of the yearSelect dropdown
+function handleYearSelectChange(params) {
```
anthonygego: Well, actually, after looking more carefully, this can probably be handled in a cleaner fashion by :

1. Adding a `data-redirect` attribute to each ` select > option` with the result of each `url_for`.
3. In the `change` listener below, change the window location to the value of `$(this).find(':selected").data("redirect")`

anthonygego (commented):
These are the last comments after code testing. There are no length specified for the strings in the models: what DBMS are you using ? Ideally, we would like to rely on MySQL/MariaDB .

SamuelVch98:
> These are the last comments after code testing. There are no length specified for the strings in the models: what DBMS are you using ? Ideally, we would like to rely on MySQL/MariaDB .

I use PostgreSQL but it's true that I should add a size limit for strings. Should I make the change straight away or can I add it in another PR?

anthonygego:
LGTM. Please rebase the next PRs to master. Thanks

PR #6, opened 2024-07-11
Feature display teacher researcher history

SamuelVch98:
- Display on the course information page the assistants assigned for the current year + history of former assistants 
- On the home page display : 

   - If you are a researcher: courses assigned to you + history of previously assigned courses and assessments 

   - If you are a teacher: the courses you are teaching this year + the courses assigned to your researchers and their assessments

### DIFF
--- a/app.py
+++ b/app.py
@@ -7,7 +7,7 @@ from user import user_bp
 from course import course_bp
 from config import config_bp
 from course_preference import course_preference_bp
-from db import db, Configuration, Organization, User, Course
+from db import db, Configuration, Organization, User, Course, Teacher
 from decorators import *
 from flask import Flask, render_template, session, request
 from enums import *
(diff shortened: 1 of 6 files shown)


app.py
```diff
@@ -46,7 +46,14 @@ def inject_configurations():
 @app.route('/')
 def index():  # put application's code here
     if session and session['logged_in']:
```
anthonygego: Is `@login_required` not applicable here ?

app.py
```diff
@@ -46,7 +46,14 @@ def inject_configurations():
...
-        return render_template("home.html")
+        user = db.session.query(User).filter_by(email=session['email']).first()
+        courses = db.session.query(Course).all()
+        teachers_course = {
```
anthonygego: Can't you use the backrefs from the Teacher model to the courses (and therefore an SQL request), instead of filtering manually in a Python for loop ?

anthonygego (changes requested):

anthonygego (approved):
LGTM

PR #8, opened 2024-07-12
Feature evaluations

SamuelVch98:
Add a page for writing a course evaluation + database storage

### DIFF
--- a/course.py
+++ b/course.py
@@ -69,7 +69,7 @@ def assign_teachers_to_course(course_id, course_year, assigned_teachers):
 @login_required
 def add_course():
     if request.method == 'GET':
-        return redirect(url_for('course.form_course'))
+        return render_template('add_course.html')
 
     form = request.form
     if not form:
@@ -100,7 +100,7 @@ def add_course():
 
         db.session.add(new_course)
         db.session.commit()
-        return redirect(url_for("course.form_course"))
+        return redirect(url_for("course.courses", current_year=year))
     except Exception as e:
         db.session.rollback()
         raise e
--- a/course.py
+++ b/course.py
@@ -1,5 +1,5 @@
 from decorators import login_required
-from db import db, User, Course, Teacher, Organization, Configuration
+from db import db, User, Course, Teacher, Organization, Evaluation
 from flask import Blueprint, render_template, flash, url_for, request, make_response, redirect, \
     Flask, jsonify, session
 from util import get_current_year
@@ -33,6 +33,7 @@ def validate_number_pattern(number):
 def form_course():
     return render_template('add_course.html')
 
+
 def validate_form_data(form, extra_fields_needed=False):
     mandatory_fields = {
         'code': validate_course_code,
@@ -62,6 +63,7 @@ def validate_form_data(form, extra_fields_needed=False):
                 return f"Missing {field}", 400
             if not validator(value):
                 return f"Invalid {field}", 400
+
 
 def assign_teachers_to_course(course_id, course_year, assigned_teachers):
     try:
@@ -132,7 +134,8 @@ def search_teachers():
     if not validate_string_pattern(search_term):
         return make_response("Invalid search term", 400)
 
-    teachers = db.session.query(User).filter(User.active == True, User.is_teacher == True, User.name.ilike(f'%{search_term}%')).all()
+    teachers = db.session.query(User).filter(User.active == True, User.is_teacher == True,
+                                             User.name.ilike(f'%{search_term}%')).all()
     results = [{'id': teacher.id, 'text': f'{teacher.name} {teacher.first_name}'} for teacher in teachers]
     return jsonify(results)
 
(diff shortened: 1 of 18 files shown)


course.py
```diff
@@ -248,3 +248,44 @@ def add_duplicate_course():
         db.session.rollback()
 
     return redirect(url_for('course.course_info', course_id=course_id, current_year=year))
+
+
+@course_bp.route('/evaluations')
+@login_required
+def evaluations():
+    user_id = request.args.get('user_id')
```
anthonygego: It seems you are confused between route and GET arguments. The route processor embeds regex matching and argument existence.

It is therefore more interesting to rely on the route processor for mandatory arguments and on the GET arguments for optional ones.

In this case, user id and current year should probably be part of the route arguments as you would to ensure the argument is specified and is correct.

course.py
```diff
@@ -248,3 +248,44 @@ def add_duplicate_course():
...
+
+@course_bp.route('/create_evaluations', methods=['POST'])
+@login_required
+def create_evaluation():
+    form = request.form
+    if not form:
+        return make_response("Problem with form request", 500)
+
+    course_id = request.form['course_id']
+    current_year = request.args.get('current_year')
```
anthonygego: Same comment here.

course.py
```diff
@@ -248,3 +248,44 @@ def add_duplicate_course():
...
+
+
+@course_bp.route('/create_evaluations', methods=['POST'])
+@login_required
+def create_evaluation():
+    form = request.form
+    if not form:
+        return make_response("Problem with form request", 500)
+
+    course_id = request.form['course_id']
```
anthonygego: Beware, you may get a malformed form and not get the expected keys. Please handle this case.

course.py
```diff
@@ -248,3 +248,44 @@ def add_duplicate_course():
...
+    current_year = request.args.get('current_year')
+    user_id = request.args.get('user_id')
+    tasks = request.form.getlist('tasks[]')
+    evaluation_hour = request.form['evaluation_hour']
+    workload = request.form['workload']
+    comment = request.form['comment']
+    second_course = True if request.form.get('second_course') == 'Yes' else False
+
+    tasks_str = ', '.join(tasks)
```
anthonygego: You may want something more robust and keep a list in a JSON (or equivalent) format. This way the code won't break even if we decide to add a task including a comma.

db.py
```diff
@@ -149,3 +149,23 @@ class CourseOrganization(db.Model):
...
+
+
+class Evaluation(db.Model):
+    __tablename__ = 'evaluation'
+
+    course_id = db.Column(db.Integer, primary_key=True)
+    course_year = db.Column(db.Integer, primary_key=True)
+    user_id = db.Column(db.Integer, db.ForeignKey('user.id'), primary_key=True)
+    task = db.Column(db.String)
+    nbr_hours = db.Column(db.String)
```
anthonygego: String ?

anthonygego (changes requested):
Use route arguments for mandatory ones and GET arguments for optional ones only. Please apply this rule in the whole project. We may already have failed to respect it.

course.py
SamuelVch98: I understand the difference better now, thank you!

course.py
```diff
@@ -249,3 +249,45 @@ def add_duplicate_course():
...
+
+    try:
+        new_evaluation = Evaluation(course_id=course_id, course_year=current_year, user_id=user_id, task=tasks,
+                                    nbr_hours=evaluation_hour, workload=workload, comment=comment,
+                                    second_course=second_course)
+        db.session.add(new_evaluation)
+        db.session.commit()
+    except Exception as e:
+        db.session.rollback()
+        raise e
```
anthonygego: This exception is propagated but not managed. Actually the form allows to submit an evaluation twice. However the database does not allow it. If one tries to submit the form twice, the exception is thrown.

course.py
```diff
@@ -249,3 +249,45 @@ def add_duplicate_course():
         db.session.rollback()
 
     return redirect(url_for('course.course_info', course_id=course_id, current_year=year))
+
+
+@course_bp.route('/evaluations/<int:user_id>/<int:current_year>')
+@login_required
+def evaluations(user_id, current_year):
+    courses = db.session.query(Course).filter_by(year=current_year).all()
```
anthonygego, line 258: Maybe the courses that were already evaluated can be filtered out, or, their evaluation can be replaced by a newer one.

course.py
```diff
@@ -249,3 +249,45 @@ def add_duplicate_course():
...
+        new_evaluation = Evaluation(course_id=course_id, course_year=current_year, user_id=user_id, task=tasks,
+                                    nbr_hours=evaluation_hour, workload=workload, comment=comment,
+                                    second_course=second_course)
+        db.session.add(new_evaluation)
+        db.session.commit()
+    except Exception as e:
+        db.session.rollback()
+        raise e
+
+    return redirect(url_for('course.evaluations', user_id=user_id, current_year=current_year))
```
anthonygego, line 303: You should add some feedback here. Was the form correctly processed ? Was there an issue ? Evaluation already submitted for this course or evaluation replaced by a newer one...

anthonygego (changes requested):

course.py
```diff
@@ -249,3 +250,55 @@ def add_duplicate_course():
...
+
+        new_evaluation = Evaluation(course_id=course_id, course_year=current_year, user_id=user_id, task=tasks,
+                                    nbr_hours=evaluation_hour, workload=workload, comment=comment,
+                                    second_course=second_course)
+        db.session.add(new_evaluation)
+        db.session.commit()
+    except Exception as e:
+        db.session.rollback()
+        flash(f'An error occurred: {str(e)}', 'danger')
+        raise e
```
```suggestion
```

anthonygego (approved):

PR #9, opened 2024-07-13
Feature role

SamuelVch98:
I've implemented the access level using a decorator and a method that checks the user type. I've tried to use the simplest system to implement this protection on access routes

### DIFF
--- a/course.py
+++ b/course.py
@@ -146,6 +146,7 @@ def course_info(course_id):
 
     all_years = db.session.query(Course).filter_by(id=course.id).distinct(Course.year).order_by(
         Course.year.desc()).all()
+
     return render_template('course_info.html', course=course, all_years=all_years, current_year=current_year)
 
 
--- a/course.py
+++ b/course.py
@@ -28,13 +28,6 @@ def validate_number_pattern(number):
     return re.match(pattern, number) is not None
 
 
-@course_bp.route('/form_course')
-@login_required
-@check_access_level('admin')
-def form_course():
-    return render_template('add_course.html')
-
-
 def validate_form_data(form, extra_fields_needed=False):
     mandatory_fields = {
         'code': validate_course_code,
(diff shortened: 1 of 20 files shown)


decorators.py
```diff
@@ -13,3 +14,21 @@ def decorated_function(*args, **kwargs):
             return f(*args, **kwargs)
 
     return decorated_function
+
+
+def check_access_level(access_level):
+    def decorator(f):
+        @wraps(f)
+        def decorated_function(*args, **kwargs):
+            if not session.get("logged_in", False):
```
anthonygego: This piece of code is probably not mandatory as you still use `login_required` decorator throughout the app. It is a good practice to remain explicit so keeping this code would check if the user is logged twice.

decorators.py
```diff
@@ -13,3 +14,21 @@ def decorated_function(*args, **kwargs):
...
+                return auth.login()
+
+            user = db.session.query(User).get(session["user_id"])
+            if not user.allowed(access_level):
+                return redirect(url_for("index"))
```
anthonygego, line 26: I'd recommend to be explicit with the user and return a permission denied notification. This way the user knows that there is something wrong with their rights in case the usage is legitimate, and doesn't suspect any bug.

db.py
```diff
@@ -32,6 +32,16 @@ def validate_active(self, key, value):
                 raise ValueError("Cannot deactivate a teacher assigned to a course.")
         return value
 
+    def allowed(self, access_level):
```
anthonygego, line 41: Maybe you should use an enum for the roles. This would be more robust at runtime. Python won't see we made a typo when using the decorator with simple strings.

anthonygego (changes requested):

course.py
```diff
@@ -28,6 +28,13 @@ def validate_number_pattern(number):
     return re.match(pattern, number) is not None
 
 
+@course_bp.route('/form_course')
+@login_required
+@check_access_level(Role.ADMIN)
+def form_course():
+    return render_template('add_course.html')
```
anthonygego: Is this endpoint useful ?

anthonygego (changes requested):

anthonygego (approved):

PR #10, opened 2024-07-16
Integrate the course assignment table according to researchers' preferences

SamuelVch98:
Integration of the handsontable table into the application. 
Ability to save the state of a table, delete data, export it as a CSV file and publish it.
Update README


templates/assignment.html
```diff
@@ -0,0 +1,65 @@
...
+            <button type="button" class="btn-close" data-bs-dismiss="toast" aria-label="Close"></button>
+        </div>
+        <div class="toast-body">
+            Placeholder for notification message
+        </div>
+    </div>
+    <div class="row">
+        <div class="col">
+            <div id="handsontable">
+                <script src="{{ url_for('static', filename='scripts/index.js') }}"></script>
```
anthonygego: You probably want to keep this file in the `static` folder, with a more specific name.

README.md
```diff
@@ -136,4 +136,11 @@
     ```
     Replace the `<tags>` with the desired information for the first admin.
 
-1. Run the IdP server and the app.
\ No newline at end of file
+1. Run the IdP server and the app.
+
+### Setting up Handsontable
```
anthonygego: Setting up...or building ? Only the `min.js` file is useful for the project.

assignment.py
```diff
@@ -0,0 +1,71 @@
...
+    current_year = get_current_year()
+    courses = db.session.query(Course).order_by(Course.quadri).all()
+    users = db.session.query(User).filter_by(admin=False, active=True).all()
+    teachers = db.session.query(Teacher).all()
+    researchers = db.session.query(Researcher).all()
+    preferences = db.session.query(PreferenceAssignment).all()
+    organizations = db.session.query(Organization).all()
+
+    data = {
+        'courses': [serialize_model(course) for course in courses if course.year == current_year],
```
anthonygego: Filtering should be done in the database request.

assignment.py
```diff
@@ -0,0 +1,71 @@
...
+
+def serialize_model(model):
+    """Converts a SQLAlchemy model object into a dictionary."""
+    return {column.name: getattr(model, column.name) for column in model.__table__.columns}
+
+
+@assignment_bp.route('/load_data', methods=['GET'])
+@login_required
+def load_data():
+    current_year = get_current_year()
```
anthonygego, line 27: You should use this to filter the data you're fetching from the database, when applicable.

assignment.py
```diff
@@ -0,0 +1,71 @@
...
+    courses = db.session.query(Course).order_by(Course.quadri).all()
+    users = db.session.query(User).filter_by(admin=False, active=True).all()
```
anthonygego: There is no need to disable admins. Moreover, some admins are actual researchers.

static/scripts/index.js
```diff
@@ -0,0 +1,513 @@
+fetch('/assignment/load_data')
+    .then(response => response.json())
+    .then(loadData => {
+        console.log(loadData);
```
anthonygego:
```suggestion

```
We probably don't want this in the repo

static/scripts/index.js
```diff
@@ -0,0 +1,513 @@
...
+
+        const {courses, users, teachers, researchers, preferences, organizations, current_year} = loadData;
+        const example = document.getElementById("handsontable");
+
+        courses.map(course => {
+            let teachersId = teachers.filter(teacher => teacher.course_id === course.id && teacher.course_year === current_year);
```
anthonygego: The year filter should be applied at the SQL side, we are only considering a year at a time here.

static/scripts/index.js
```diff
@@ -0,0 +1,513 @@
...
+            let teachersName = teachersId.map(teacher => {
+                let teacherUser = users.find(user => user.id === teacher.user_id);
```
anthonygego: You may want to generate dictionaries and not lists in `assignment.py` to avoid a find operation and access the teacher data structure more efficiently.

static/scripts/index.js
```diff
@@ -0,0 +1,513 @@
...
+                return teacherUser ? teacherUser.name + " " + teacherUser.first_name : '';
+            });
+            course.assigned_teachers = teachersName.join(', ');
+        });
+
+        const fixedHeaders = [
+            "Fgs",
```
anthonygego: FGS is the UCLouvain numerical id. In thi app the user id is an autoincremented private integer that won't match. You shouldn't refer it as to FGS` If you want to provide directory information to the administrator, maybe the email would be more useful, to distinguish homonyms.

static/scripts/index.js
```diff
@@ -0,0 +1,513 @@
...
+            "Total Load",
+            "Load Q1",
+            "Load Q2",
+        ];
+
+        const fixedRowsText = [
+            "Id",
+            "Quadri",
+            "Acads",
+            "Nbr Student 22/23",
```
anthonygego: I guess that `22/23` means the academic year *2022-2023*. This should probably be removed. We can consider the saved value topical.

static/scripts/index.js
```diff
@@ -0,0 +1,513 @@
...
+                row.researchers.fgs = "";
+                row.researchers.name = "";
+                row.org = "";
+                row.charge1 = "";
+                row.charge2 = "";
+                row.charge3 = "";
+                row.check = "";
+
+                courses.forEach(course => {
+                    row[course.code] = properties[i] !== 'tutors' ? course[properties[i]] : 0;
```
anthonygego: If properties are keys of `course`, you can probably just generate the  `properties` list from the `course` keys and avoid hardcoding them at several locations.

static/scripts/index.js
```diff
@@ -0,0 +1,513 @@
...
+            return row;
+        }
+
+        function userRowsData() {
+            const rows = [];
+            const researcherUsers = users.filter(user => user.is_researcher === true);
+
+            researcherUsers.forEach(user => {
+                const row = buildRow(user, true);
+                const emptyRow = buildRow(user, false);
```
anthonygego: You should add a comment to explain why the two rows, otherwise we'll forget with time.

static/scripts/index.js
```diff
@@ -0,0 +1,513 @@
...
+
+                const matchingAssistant = researchers.find(researcher => researcher.user_id === user.id);
+                const assistantOrg = organizations.find(org => org.id === user.organization_id);
```
anthonygego: Same, you can gain in performance by generating dictionaries instead of lists.

static/scripts/index.js
```diff
@@ -0,0 +1,513 @@
...
+
+                row.org = assistantOrg ? assistantOrg.name : "";
+                row.charge1 = matchingAssistant ? (users.find(sup => sup.id === user.supervisor_id)?.name ?? "") : "";
```
anthonygego: Same, make `users` a dictionary.

static/scripts/index.js
```diff
@@ -0,0 +1,513 @@
...
+                row.charge2 = matchingAssistant ? matchingAssistant.max_loads : 0;
+                row.charge3 = 0;
+                row.check = 0;
+
+                emptyRow.org = "";
+                emptyRow.charge1 = "";
+                emptyRow.charge2 = "";
+                emptyRow.charge3 = "";
+                emptyRow.check = "";
```
anthonygego: Again here please make explicit by a comment why we do this. By the way, is there no mean to express the row data as a list, as all the values are empty ?

static/scripts/index.js
```diff
@@ -0,0 +1,513 @@
...
+
+                const researcher = researchers.find(researcher => researcher.user_id === user.id);
```
anthonygego: Same, make `researchers` a dictionary.

static/scripts/index.js
```diff
@@ -0,0 +1,513 @@
...
+                const userPrefs = preferences.filter(pref => pref.researcher_id === researcher.id && pref.course_year === current_year);
```
anthonygego: Filtering by year should be done in the SQL request.

static/scripts/index.js
```diff
@@ -0,0 +1,513 @@
...
+
+                let pos = 1;
+                courses.forEach(course => {
+                    const isPref = userPrefs.find(pref => pref.course_id === course.id && pref.course_year === current_year);
```
anthonygego: Same, dictionary for `userPrefs` and no year filtering here.

static/scripts/index.js
```diff
@@ -0,0 +1,513 @@
...
+        }
+
+        const columns = getCourseColumns();
+
+        let data = fixedRows.concat(userRows);
+        const nbrLines = data.length - 1;
+        const nbrCols = columns.length - 1;
+
+        const startRow = lenFixedRowsText;
+        const startCol = 1;
```
anthonygego: Do you actually need this definition ?

static/scripts/index.js
```diff
@@ -0,0 +1,513 @@
...
+        const endRow = nbrLines;
+        const endCol = lenFixedHeaders - 1;
+
+        const mergeCellsSettings = generateMergeCellSettings(startRow, startCol, endRow, endCol);
+
+        function generateMergeCellSettings(startRow, startCol, endRow, endCol) {
```
anthonygego: Do you actually need to define a function for this ?

static/scripts/index.js
```diff
@@ -0,0 +1,513 @@
...
+                            //Boolean to determine whether a cell is updated or empty
+                            let isDeleted = false;
+
+                            //Update total course assistants
+                            if (oldValue !== null && newValue === null) {
+                                isDeleted = true;
+                            }
+
+                            const colInfos = this.getDataAtProp(prop);
+                            let nbrAssistants = colInfos[5];
```
anthonygego: I realize that specific column indexes are put in the code as is. Maybe it would be more maintenable to have an enum with the column name and indexes somewhere to specify the column name instead of its index. 

This would ease column addition/deletion in the future as well updating the code if a column position is changed.

static/scripts/index.js
```diff
@@ -0,0 +1,513 @@
...
+                    },
+                },
+                {
+                    row: 5,
+                    col: 6,
+                    bottom: {
+                        width: 2,
+                        color: 'black'
+                    }
+                },
```
anthonygego: You can probably compact this snippet a little bit. The expanded dict is harder to read than it would be in a more compacted form.

static/scripts/index.js
```diff
@@ -0,0 +1,513 @@
+fetch('/assignment/load_data')
+    .then(response => response.json())
+    .then(loadData => {
+        console.log(loadData);
+
+        const {courses, users, teachers, researchers, preferences, organizations, current_year} = loadData;
+        const example = document.getElementById("handsontable");
```
anthonygego: `example` is not the ideal constant name. However, you can simply put the selector where the constant is used.

static/scripts/index.js
```diff
@@ -0,0 +1,513 @@
...
+            afterFilter(conditionsStack) {
+                const filtersPlugin = this.getPlugin('filters');
+                const filtersRowsMap = filtersPlugin.filtersRowsMap;
+
+                // Exclude fixed lines from the filter
+                for (let i = 0; i < lenFixedRowsText; i++) {
+                    filtersRowsMap.setValueAtIndex(i, false);
+                }
+            },
+            licenseKey: "non-commercial-and-evaluation",
```
anthonygego: Do we really need to specify this ? :eyes:

static/scripts/index.js
```diff
@@ -0,0 +1,513 @@
...
+            },
+            afterFilter(conditionsStack) {
+                const filtersPlugin = this.getPlugin('filters');
+                const filtersRowsMap = filtersPlugin.filtersRowsMap;
+
+                // Exclude fixed lines from the filter
+                for (let i = 0; i < lenFixedRowsText; i++) {
+                    filtersRowsMap.setValueAtIndex(i, false);
+                }
+            },
```
anthonygego: Is this piece of code really used ? It seems a bit confusing to me.

static/scripts/index.js
```diff
@@ -0,0 +1,513 @@
...
+                    resetDataLocally();
+                    updateToastContent('Data cleared');
+                    toastNotification.show();
+                    setTimeout(function () {
+                        location.reload();
+                    }, 1500);
+                })
+                $('#button-publish-assignments').click(async function () {
+                    const slicedData = data.slice(lenFixedRowsText);
+                    console.log("Sliced Data", slicedData);
```
anthonygego:
```suggestion

```
Probably to be removed

static/scripts/index.js
```diff
@@ -0,0 +1,513 @@
...
+                    const result = [];
+
+                    for (let i = 0; i < slicedData.length; i += 2) {
+                        const user_row = slicedData[i];
+                        const course_row = slicedData[i + 1];
+
+                        const userData = {
+                            user_id: user_row.researchers.fgs,
+                            load_q1: user_row.charge3,
```
anthonygego: Maybe it would be nice to have the names (q1/charge3) matching.

templates/assignment.html
```diff
@@ -0,0 +1,65 @@
...
+        </div>
+    </div>
+    <br>
+    <div class="toast" id="toast-notification" role="alert" aria-live="assertive" aria-atomic="true">
+        <div class="toast-header">
+            <strong class="me-auto">Notification</strong>
+            <button type="button" class="btn-close" data-bs-dismiss="toast" aria-label="Close"></button>
+        </div>
+        <div class="toast-body">
+            Placeholder for notification message
```
anthonygego: Does this `<div/>`can remain empty ? Is it actually shown ?

anthonygego (changes requested):

templates/assignment.html
SamuelVch98: The ‘function updateToastContent(message)’ function updates the content of the toast body, so this is necessary

static/scripts/index.js
SamuelVch98: If I remove this property, this message appears below the table: 

‘The licence key for Handsontable is missing. Use the key you purchased to activate the product. Alternatively, you can activate Handsontable for non-commercial use by passing the key: ‘non-commercial-and-evaluation’. For more information, consult the documentation or contact us at support@handsontable.com. ’

static/scripts/index.js
SamuelVch98: Yes, it's used to filter data. Handsontable does not take merged rows into account when sorting. So I have to do it manually, I retrieve which values are requested and in which column. Then I go through the rows of user data and check whether the value matches the one requested to retrieve the row id. Then I create a new condition, this time based on the id, to get the merged rows.

static/scripts/index.js
SamuelVch98: I don't know how to simplify this code, because I don't need all the keys in the course and the order of the attributes is important

static/scripts/assignment_table.js
```diff
@@ -0,0 +1,510 @@
...
+        const fixedRowsText = [
+            "Id",
+            "Quadri",
+            "Acads",
+            "Nbr Students",
+            "Assistant",
+            "Total for now",
+        ];
+
+        const lenFixedRowsText = fixedRowsText.length
```
anthonygego: You should check the whole file for missing semicolons. While the browser JS engine can fix this on runtime, it remains very bug prone.

static/scripts/assignment_table.js
```diff
@@ -0,0 +1,510 @@
...
+        //Create all course header
+        const allHeaders = fixedHeaders.concat(coursesHeaders);
+
+        function fixedRowsData() {
+            let rows = [];
+
+            const allProperties = Object.keys(courses[0]);
+            const requiredProperties = ['id', 'quadri', 'assigned_teachers', 'nbr_students', 'nbr_teaching_assistants', 'tutors'];
+            const properties = allProperties.filter(prop => requiredProperties.includes(prop));
+            console.log("Properties", properties);
```
anthonygego:
```suggestion
```

static/scripts/assignment_table.js
```diff
@@ -0,0 +1,510 @@
...
+                const slicedData = data.slice(lenFixedRowsText);
+                const result = [];
+
+                for (let i = 0; i < slicedData.length; i += 2) {
+                    const user_row = slicedData[i];
+                    const course_row = slicedData[i + 1];
+
+                    const userData = {
+                        user_id: user_row.researchers.id,
+                        load_q1: user_row.charge3,
```
anthonygego: Did you forget to update the column id `charge3` here ?

anthonygego (changes requested):

anthonygego:
Can you rebase ?

static/scripts/assignment_table.js
```diff
@@ -0,0 +1,513 @@
...
+            afterChange: function (changes) {
+                if (changes) {
+                    changes.forEach(([row, prop, oldValue, newValue]) => {
+                        let col = this.propToCol(prop);
+                        if ((col >= lenFixedHeaders && row >= lenFixedRowsText) && (row % 2 === 1)) {
+                            //Boolean to determine whether a cell is updated or empty
+                            let isDeleted = false;
+
+                            //Update total course assistants
+                            if (oldValue !== null && newValue === null) {
```
anthonygego: We need something more robust here... in pratice this is not really usable...

In practice, whenever we change the content of the cell to anything else than null, the load is incremented. The event is triggered once we get and release the focus on the cell.

This means you can:
- indefinetely increase the load by getting in and out the cell.
- whatever you put in the cell, you increase the load by one.
- you cannot simply add a load of 2, and then decrease to a load of 0.
- pressing delete multiple times on an empty cell will act as an increase.

It's probably more interesting to rely on the numerical value specified in the cell

static/scripts/assignment_table.js
```diff
@@ -0,0 +1,513 @@
...
+                    if (col === ColumnIndices.LOAD_Q1 || col === ColumnIndices.LOAD_Q2) {
+                        const total_load = rowValue[ColumnIndices.TOTAL_LOAD];
+                        const load_q1 = rowValue[ColumnIndices.LOAD_Q1];
+                        const load_q2 = rowValue[ColumnIndices.LOAD_Q2];
+
+                        //First case: load Q1 + Q2 = total load
+                        //Second case: load Q1 or Q2 = half of total load
+                        if (load_q1 + load_q2 > total_load) {
+                            TD.style.backgroundColor = 'red';
+                        } else if ((total_load === 4 || total_load === 2) && (total_load / value <= 2 || load_q1 + load_q2 === total_load)) {
```
anthonygego: Constants should probably be defined in a more textual way somewhere.

anthonygego (changes requested):
It seems  that we can change the preferences line and put whatever we want in the cells, and then save the table and loose the actual prferences...

Also, I think that you should add some short notice on how to use the table: which cells need to be modified, how....

anthonygego:
Can you fix conflicts and rebase ?

course_preference.py
```diff
@@ -41,14 +40,15 @@ def save_preference():
     if researcher is None:
         return make_response("User is not a researcher", 500)
 
-    new_course_ids = {preference['course_id'] for preference in preferences}
-    delete_old_preferences(researcher.id, new_course_ids, current_year)
+    delete_old_preferences(researcher.id, current_year)
 
+    rank = 0
```
anthonygego, line 45: It seems this piece of code comes from #42. We'll wait for it to be merged before proceeding further with this one.

static/scripts/assignment_table.js
```diff
@@ -0,0 +1,482 @@
...
+                    filename: 'ICTM-CSV-file_[YYYY]-[MM]-[DD]',
+                    mimeType: 'text/csv',
+                    rowDelimiter: '\r\n',
+                    rowHeaders: true
+                });
+                updateToastContent('Data exported to CSV');
+                toastNotification.show();
+            })
+            $('#button-create-assignments').click(function () {
+                storeDataLocally(data);
```
anthonygego: Do we actually want this ? Storing this in the local storage only allows the local browser to load saved data. This means:

- an admin currently working on assignments should keep using the same browser until assignments are published
- if an admin save their own assignments, another admin cannot get the current state of assignments and go on until the data are ready for publishment.

Data should be saved in remote database. And can only be viewed by teachers and researchers once they are published.

assignment.py
```diff
@@ -0,0 +1,74 @@
...
+    """Converts a SQLAlchemy model object into a dictionary."""
+    return {column.name: getattr(model, column.name) for column in model.__table__.columns}
+
+
+@assignment_bp.route('/load_data', methods=['GET'])
+@login_required
+def load_data():
+    current_year = get_current_year()
+    courses = (db.session.query(Course).filter_by(year=current_year).order_by(Course.quadri).all())
+    users = db.session.query(User).filter_by(active=True).all()
```
anthonygego: Beware: if you save assignments and then de-activate a user, their data won't be found when reloading the assignment table. You should check for this.

static/scripts/assignment_table.js
```diff
@@ -0,0 +1,482 @@
...
+            const cellsMeta = localStorage.getItem('cellsMeta');
+            return cellsMeta ? JSON.parse(cellsMeta) : null;
+        }
+
+        function resetDataLocally() {
+            localStorage.clear();
+        }
+
+        // Verify if the table is modified locally
+        retrieveDataLocally();
```
anthonygego: Please either:

- remove the function definition as it is used once and alter the value of a variable out of its scope (`data`)
- make the function return the loaded value.

static/scripts/assignment_table.js
```diff
@@ -0,0 +1,482 @@
...
+                            const quadri = colInfos[RowIndices.QUADRI];
+
+                            let loadKey = quadri === 1 ? ColumnIndices.LOAD_Q1 : ColumnIndices.LOAD_Q2;
+                            let loadValue = rowInfos[loadKey];
+
+                            // Old non-empty value and new empty value: loadValue-- and nbrAssistants--.
+                            // Old non-empty value and new non-empty value: loadValue++ and nbrAssistants++.
+                            // Old null value and new non-empty value: loadValue++ and nbrAssistants++.
+                            // Old non-null value and new null value: loadValue-- and nbrAssistants--.
+                            if (oldValue !== newValue) {
```
anthonygego: We still need a small doc for this in the UI. It won't be clear for user that whatever the cell content is corresponds to *assigned* and that an empty cell corresponds to *not assigned*.

Another nice and valid solution would be to catch double click for instance and *check* or *uncheck* the cell.

static/scripts/constants.js
```diff
@@ -0,0 +1,45 @@
...
+
+    export const lenFixedRowsText = fixedRowsText.length; // Number of fixed rows
+    export const lenFixedHeaders = fixedHeaders.length; // Number of fixed headers
+    export  const borderStyle = {width: 2, color: 'black'}; // Border style for the table
+
+    export const requiredProperties = ['id', 'quadri', 'assigned_teachers', 'nbr_students', 'nbr_teaching_assistants']; // Required properties for the table
+
+    export const taLoad = 4; // Load of a teaching assistant
+    export const phdLoad = 2; // Load of a phd student
+    export const postDocLoad = 1; // Load of a post doc
```
anthonygego, line 45: This is already defined in `enums.py`. You should rather inject them dynamically from the template instead of duplicating definitions.

anthonygego (changes requested):

db.py
```diff
@@ -190,6 +190,14 @@ class PublishAssignment(db.Model):
     )
 
 
+class SaveAssignment(db.Model):
```
anthonygego: Relational database systems are there to store related pieces of data and ensure the whole data integrity through operations. 

By storing various data, including their relations, in large TEXT/BLOB/JSON columns, you're breaking the embedded integrity check system and make the use of such a database system irrelevant, as it is not capable of parsing the column content.

There should be no difference between an assignment before and after publishing, except it is shown to users. Assignments can be saved in a unique table then, whether they are still part of a draft or not.

anthonygego (changes requested):

db.py
```diff
@@ -172,6 +173,51 @@ class CourseOrganization(db.Model):
...
+            ['course.id', 'course.year']
+        ),
+    )
+
+
+class Comment(db.Model):
+    __tablename__ = 'comment'
+
+    id = db.Column(db.Integer, primary_key=True)
+    row = db.Column(db.Integer, nullable=False)
```
anthonygego: I'm totally confused here. Can't we simply associate the comment to the assignment, which corresponds to a cell in practice ?

Saving cell column and line won't work if the dataset change (courses or researchers).

db.py
```diff
@@ -172,6 +173,51 @@ class CourseOrganization(db.Model):
     )
 
 
+class Assignment(db.Model):
```
anthonygego: In all cases we don't need more than a unique table for assignments: comments and, potentially, draft fields included. A system-wide draft flag can be stored in the Configuration model.

anthonygego (changes requested):

db.py
SamuelVch98: I don't understand why we should use the configuration table. It would mean that there would have to be a link between configuration and the line of assignments. 

That's why I'd prefer to keep the configuration table to store all the information related to the years and the assignment table would serve as a link between the common assignment lines but also to store the status of an assignment.

anthonygego: >  I don't understand why we should use the configuration table. 

We shouldn't, we can. There is one process per academic year. And in this process the assignments are either draft or published.

> It would mean that there would have to be a link between configuration and the line of assignments.

There exists an implicit relation between an academic year and the assignments made for that year. If that academic year doesn't exist, there is no reason for the assignments to exist.

The current table name, which, by the way, only store if the academic year is the current one or not, should not restrict you in implementation choices. That table could also be named `Process`, `Period`, `Year`...

That process is very similar to the fiscal one. Each year, a fiscal year begins and lasts a given period of time, according to the state, with potentially specific rules/adjustments for that year and several steps to be achieved.

SamuelVch98: Firstly, I understand the link between the configuration table and the assignment table. This would mean that for each year, there would be two configurations to store draft/published assignments.

Example for the year 2024: 

 I would therefore have two configurations, one where the assignments are drafts and one for publications. So when I publish new assignments for a configuration, the previous assignments will be replaced.

AlexandreDoneux: First, I think we should rename the configuration table "Year" (or something else Anthony proposed) so it is more clear. 

I'm not sure what you are trying to explain with your example. I feel like you just need a configuration per year and the info of it being a draft inside the configuration table. You would then have an assignment table (currently AssignmentLine) where we have an entry per link between user and course-year.

db.py
AlexandreDoneux: If there is only one comment per assignment then we should definitely have it stored in the assignment entry. 

Yes, the comment will have to be generated using the row and column info at the moment we generate the table.
You should always design an app with keeping in mind the fact data and data structure can change.

static/scripts/assignment_table.js
```diff
@@ -0,0 +1,481 @@
...
+
+                        if ((col >= lenFixedHeaders && row >= lenFixedRowsText) && (row % 2 === 1)) {
+                            const colInfos = this.getDataAtProp(prop);
+                            const quadri = colInfos[RowIndices.QUADRI];
+                            let loadKey = quadri === 1 ? ColumnIndices.LOAD_Q1 : ColumnIndices.LOAD_Q2;
+
+                            //Count the number of assignments to determine the value of loads
+                            let researcherRow = this.getDataAtRow(row).slice(lenFixedHeaders);
+                            let loadValue = researcherRow.filter(value => value !== null && value !== '').length;
+                            this.setDataAtCell(row - 1, loadKey, loadValue);
```
AlexandreDoneux: There is a big problem the way the total load for a researcher behaves. When assigning a course, you count all the non selected cells and apply the value to the quadri load of the course you just assigned. For example, when you assign a course from Q2 to a researcher, all his assignations are counted (Q1 and Q2) and the value is set to his Q2 total load.

You should have a way to only count the assignations from the correct semester (quadri).

[Screencast from 11-07-24 15:31:14.webm](https://github.com/user-attachments/assets/75f147a2-f7ca-4e03-a923-55c99a325bd0)

The same applies when removing assignations.

AlexandreDoneux: Also, when indicating teacher names you only use the last name (here numbers). If multiple teachers have the last first name there is no way to differentiate them (yes, that happens). Either write the full names or add the first letter of their name. The second would not be perfect but would greatly reduce the risk.

static/scripts/assignment_table.js
```diff
@@ -0,0 +1,481 @@
...
+                th.style.fontWeight = 'bold';
+                th.style.height = '100px';
+                th.style.lineHeight = '100px';
+
+                if (col > lenFixedHeaders - 1) {
+                    let colData = this.getDataAtCol(col);
+                    if (colData[RowIndices.TOTAL_ASSISTANT_NOW] >= colData[RowIndices.ASSISTANTS]) {
+                        th.style.backgroundColor = 'green';
+                    }
+                }
```
AlexandreDoneux: Your check on the number of assignations per course is not complete because you only check if that number is lower or equal to the desired value. You don't check if it goes over (red).

Also, you don't do that check if there is no desired number of assigned researcher. The title remains green all the time. I don't know what is best : giving a number at course creation or just leaving the title white if no number is given. What do you think @anthonygego @luciledierckx

AlexandreDoneux (changes requested):

static/scripts/assignment_table.js
```diff
@@ -0,0 +1,481 @@
...
+                await saveAssignment(true);
+            });
+
+            $('#button-clear-assignments').click(function () {
+                const newUsersRow = userRowsData(true);
+                data = fixedRows.concat(newUsersRow);
+                table.loadData(data);
+                updateToastContent('Data cleared');
+                toastNotification.show();
+            });
```
AlexandreDoneux: The "clear assignment" button does clear the selections but not the values of the total per course, as well as put the title at it's default color.  

image

AlexandreDoneux: It is not a critical problem but it is something that should not be in a finished product.

static/scripts/assignment_table.js
SamuelVch98: If there is no desired number of researchers assigned, the value is set to 0 by default. Would you like me to set the header to ‘’white‘’ if the value is the default?

AlexandreDoneux: I feel that in the normal use of the application, we should define a number of researchers per course. Obviously a person creating a course might not now exactly how many are needed at first. And that is why we handle the case where none are given. 

I was thinking we might use a way to clearly indicate that desired number is not defined for a course. For example, ith a title in light purple or something similar. It would be a way of saying "Be advised, you didn't define a desired number of researcher !".
That additional color would prevent the case where we confuse the courses that are not fully assigned and those with no desired number of researchers.

I will discuss this possibility with the others.

AlexandreDoneux:
As discussed, you should indicate clearly in the table when a course has no number of researcher defined

AlexandreDoneux (commented):
As discussed, you should indicate clearly in the table when a course has no number of researcher defined

AlexandreDoneux (changes requested):
We want a course that is not fully assigned to be white. Right now it becomes red.
Also, a course with too many researchers assigned stays green instead of turning red.

static/scripts/assignment_table.js
```diff
@@ -0,0 +1,490 @@
...
+
+                let colData = this.getDataAtCol(col);
+                if (col > lenFixedHeaders - 1) {
+                    if (colData[RowIndices.ASSISTANTS] === 0) {
+                        th.style.backgroundColor = '#E1BEE7';
+                    }
+                    else if (colData[RowIndices.TOTAL_ASSISTANT_NOW] >= colData[RowIndices.ASSISTANTS]) {
+                        th.style.backgroundColor = 'green';
+                    }
+                }
```
AlexandreDoneux: Where is the red gone ? 

Now it is white when not enough researchers are assigned and turns green when enough are assigned. But if too many are assigned the titles does not change to red... Also the red color is missing from the legend.

AlexandreDoneux (changes requested):

static/scripts/assignment_table.js
SamuelVch98: Sorry, it's my fault, I misunderstood your last comment.

templates/assignment.html
```diff
@@ -0,0 +1,130 @@
...
+                                    <ul class="list-unstyled">
+                                        <li><span class="d-inline-block bg-success"
+                                                  style="width: 16px; height: 16px; margin-right: 8px;"></span> Desired
+                                            number of assistants reached
+                                        </li>
+                                        <li><span class="d-inline-block bg-danger"
+                                                  style="width: 16px; height: 16px; margin-right: 8px;"></span>
+                                            Additional
+                                            researchers needed
+                                        </li>
```
AlexandreDoneux: "Additional researchers needed" needs to be "too many researchers assigned"

AlexandreDoneux (changes requested):
Looks good to me other than that last change in the legend. @anthonygego, does it look good ?

static/img/assigned_cell.JPG
anthonygego, line 1: Many filesystems are case-sensitive. Links refer to 'assigned_cell.jpg` but the file is named `assigned_cell.JPG`. The file is not served using common Unix filesystems.

assignment.py
```diff
@@ -0,0 +1,118 @@
...
+                    load_q1=load_q1, load_q2=load_q2, position=position, comment=comment
+                ))
+
+                if not is_draft:
+                    assignments_to_add.append(AssignmentPublished(
+                        course_id=course_id, course_year=current_year, researcher_id=researcher_id,
+                        load_q1=load_q1, load_q2=load_q2, position=position, comment=comment
+                    ))
+
+    db.session.bulk_save_objects(assignments_to_add)
```
anthonygego: For me this is OK for now but this method is actually documented as a legacy feature : https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.bulk_save_objects

assignment.py
```diff
@@ -0,0 +1,118 @@
...
+    for item in data.get('data'):
+        user_data = item.get('userData')
+        course_data = item.get('courseData')
+
+        if course_data and user_data:
+            researcher_id = user_data.get('researcher_id')
+            load_q1 = user_data.get('load_q1')
+            load_q2 = user_data.get('load_q2')
+            for course_id, properties in course_data.items():
+                position = properties.get('position')
```
anthonygego: You should make sure that the data format are correct to avoid trying to add incorrect data type is database (string instead of integers for instance). You can also manage this and other cases using try-except clauses, as long as the user is informed of the error and data are not partially saved.

assignment.py
```diff
@@ -80,39 +80,55 @@ def publish_assignments():
...
+                            course_id=course_id, course_year=current_year, researcher_id=researcher_id,
+                            load_q1=load_q1, load_q2=load_q2, position=position, comment=comment
+                        ))
+                        if not is_draft:
+                            assignments_to_add.append(AssignmentPublished(
+                                course_id=course_id, course_year=current_year, researcher_id=researcher_id,
+                                load_q1=load_q1, load_q2=load_q2, position=position, comment=comment
+                            ))
+                    except (ValueError, TypeError) as e:
+                        return jsonify({"error": f"Invalid data for researcher_id {researcher_id} course_id {course_id}: {str(e)}"}), 400
```
anthonygego: Can you just return the researcher and course name so that the admin can locate where the invalid data are in the table ?

assignment.py
```diff
@@ -0,0 +1,135 @@
...
+        db.session.commit()
+
+        assignments_to_add = []
+
+        for item in data.get('data', []):
+            try:
+                user_data = item.get('userData')
+                course_data = item.get('courseData')
+
+                if not course_data or not user_data:
```
anthonygego: If there is no assignment made for a researcher (empty line), `course_data` is empty and a `ValueError` is thrown: not sure this is an expected behaviour.

Besides, there would still be handy to know where the error is. In practice there will be more than a hundred lines in this table.

anthonygego (approved):

PR #11, opened 2024-07-22
Filter on column table

SamuelVch98:
Addition of a column filter to the course and user tables + common method to avoid repetition

### DIFF
--- a/static/scripts/util.js
+++ b/static/scripts/util.js
@@ -111,10 +111,22 @@ $(document).on('click', '#filterIcon', function (e) {
 });
 
 $(document).on('keyup', '#filterRow input', function () {
+    // Get the index of the column from the data-column attribute of the input
     let column = $(this).data('column');
+
+    // Get the value entered in the input and convert it to lowercase
     let value = $(this).val().toLowerCase();
-    $('table tbody tr').filter(function () {
-        $(this).toggle($(this).find('td').eq(column).text().toLowerCase().indexOf(value) > -1);
+
+    // Iterate through each row in the table body
+    $('table tbody tr').each(function () {
+        // Get the text content of the cell at the specified column index and convert it to lowercase
+        let cellText = $(this).find('td').eq(column).text().toLowerCase();
+
+        // Check if the cell text contains the searched value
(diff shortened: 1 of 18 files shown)


static/scripts/util.js
```diff
@@ -58,3 +58,32 @@ $(document).on('click', '.remove-tag', function (e) {
...
+/*
+$('#filterRow input').on('keyup', function () {
+    let column = $(this).data('column');
+    let value = $(this).val().toLowerCase();
+    $('table tbody tr').filter(function () {
+        $(this).toggle($(this).find('td').eq(column).text().toLowerCase().indexOf(value) > -1);
+    });
+});
+ */
+
```
anthonygego: Choose between the two but do not commit commented code

anthonygego (changes requested):

static/scripts/util.js
```diff
@@ -105,4 +105,15 @@ $(document).on('click', '.remove-tag', function (e) {
...
 
+$(document).on('click', '#filterIcon', function (e) {
+    e.preventDefault();
+    $('#filterRow').toggle();
+});
 
+$(document).on('keyup', '#filterRow input', function () {
+    let column = $(this).data('column');
+    let value = $(this).val().toLowerCase();
+    $('table tbody tr').filter(function () {
```
anthonygego:
```suggestion
    $('table tbody tr').each(function () {
```

as you don't seem to use the filtered list.

static/scripts/util.js
```diff
@@ -105,4 +105,15 @@ $(document).on('click', '.remove-tag', function (e) {
...
+        $(this).toggle($(this).find('td').eq(column).text().toLowerCase().indexOf(value) > -1);
```
anthonygego: Please add a little comment here, or split the line into multiple ones, to better understand you fetch the `column`-th column content, set it to lower case and check if it contains the searched value, then only toogle the column if this is the case.

anthonygego (changes requested):

anthonygego (approved):
LGTM

Issue #14, opened 2024-07-26
[courses] filtering with no organizations


anthonygego:
@SamuelVch98  If you apply an organization filter on the course list, courses with no organization code set won't be shown anymore, whether you click on check all or uncheck all.

3 possibilities:
- Do not allow a course without organization code
- Do allow an course with no organization code to be shown when "check all", but it may be confusing if we click on all of them individually
- Do implement a "remove filter" option.

By the way, the course-dedicated code could be used for both courses and users, maybe it's worth deleting this `if` statement:

https://github.com/UCL-INGI/ictm-teaching/blob/d99b2c535f6440830533e422572f9c1cb3a116a9/static/scripts/util.js#L81-L88

PR #15, opened 2024-07-27
Issues #14 filtering with no organizations

SamuelVch98:
Remove unnecessary conditions for sorting and do not authorise a course without organisation


course.py
```diff
@@ -65,6 +65,11 @@ def assign_teachers_to_course(course_id, course_year, assigned_teachers):
         raise e
 
 
+def count_org(orgs):
+    if len(orgs) == 0:
+        return "Please select at least one organization", 400
```
anthonygego: Maybe you can use flash cards as you did for other error messages, so that we still land on a usable page once the error is triggered. 

There is no need to return a HTTP 400 code. As long as your form allows the user to submit such content, the server should not return any error code. However, the user should indeed be warned of the missing field.

anthonygego (changes requested):

course.py
SamuelVch98: Yes its better my bad

course.py
```diff
@@ -65,6 +65,10 @@ def assign_teachers_to_course(course_id, course_year, assigned_teachers):
         raise e
 
 
+def count_org(orgs):
+    return len(orgs) == 0
+
+
```
```suggestion
```

course.py
```diff
@@ -87,6 +91,9 @@ def add_course():
     language = request.form['language']
 
     organization_ids = request.form.getlist('organization_code[]')
+    if count_org(organization_ids):
```
```suggestion
    if not len(organization_ids):
```

course.py
```diff
@@ -167,6 +174,9 @@ def update_course_info():
     nbr_monitor_students = request.form['nbr_monitor_students']
     assigned_teachers = request.form.getlist('assigned_teachers[]')
     organisation_code = request.form.getlist('organization_code[]')
+    if count_org(organisation_code):
```
```suggestion
    if not len(organisation_code):
```

course.py
```diff
@@ -239,6 +249,9 @@ def add_duplicate_course():
     course_id = request.form['course_id']
     assigned_teachers = request.form.getlist('assigned_teachers[]')
     organisation_code = request.form.getlist('organization_code[]')
+    if count_org(organisation_code):
```
```suggestion
    if not len(organisation_code):
```

templates/toast_check_orgs.html
```diff
@@ -0,0 +1,29 @@
+<div class="toast-container top-0 start-50 translate-middle-x">
```
anthonygego: Can't you just re-use the code already used through the app ?

anthonygego (changes requested):

anthonygego (approved):
LGTM

PR #16, opened 2024-07-29
Issues#13 add length to VARCHAR/String

SamuelVch98:
I've added a maximum size for fields of type String, trying to determine them clearly


db.py
```diff
@@ -16,9 +16,9 @@ class Role(Enum):
 class User(db.Model):
     __tablename__ = 'user'
     id = db.Column(db.Integer, primary_key=True)
-    name = db.Column(db.String, nullable=True)
-    first_name = db.Column(db.String, nullable=True)
-    email = db.Column(db.String, nullable=True, unique=True)
+    name = db.Column(db.String(30), nullable=True)
+    first_name = db.Column(db.String(30), nullable=True)
+    email = db.Column(db.String(100), nullable=True, unique=True)
```
```suggestion
    name = db.Column(db.String(100), nullable=True)
    first_name = db.Column(db.String(100), nullable=True)
    email = db.Column(db.String(200), nullable=True, unique=True)
```

Don't be stingy: You can set a length of 100 for both so that there won't be any issue in case of aristocratic names.

db.py
```diff
@@ -53,10 +53,10 @@ class Course(db.Model):
     __tablename__ = 'course'
     id = db.Column(db.Integer, primary_key=True, autoincrement=True)
     year = db.Column(db.Integer, primary_key=True)
-    code = db.Column(db.String)
-    title = db.Column(db.String)
+    code = db.Column(db.String(10))
+    title = db.Column(db.String(50))
```
```suggestion
    title = db.Column(db.String(100))
```

This may grow very quickly...

db.py
```diff
@@ -174,9 +174,9 @@ class Evaluation(db.Model):
     course_year = db.Column(db.Integer, primary_key=True)
     user_id = db.Column(db.Integer, db.ForeignKey('user.id'), primary_key=True)
     task = db.Column(db.JSON, nullable=False)
-    nbr_hours = db.Column(db.String, nullable=False)
-    workload = db.Column(db.String, nullable=False)
-    comment = db.Column(db.String)
+    nbr_hours = db.Column(db.String(10), nullable=False)
+    workload = db.Column(db.String(10), nullable=False)
+    comment = db.Column(db.String(100))
```
```suggestion
    comment = db.Column(db.String(500))
```

500 chars is not too much for a comment zone.

anthonygego:
Can you rebase ?

anthonygego (approved):

PR #17, opened 2024-07-29
Issues#12 no permission check on profile edition

SamuelVch98:
I've added a condition to the decorator that checks whether the id received is equal to the id of the user running. If it isn't, we're sent back to home + error message.


decorators.py
```diff
@@ -16,16 +16,24 @@ def decorated_function(*args, **kwargs):
...
             user = db.session.query(User).get(session["user_id"])
-            if not user.allowed(access_level):
+
+            # Check if the user has not the required role
+            if not any(user.allowed(role) for role in allowed_roles):
+                flash("Permission denied. You do not have access to this page.", "error")
+                return redirect(url_for("index"))
+
+            # Check if the user is trying to access a profile that is not his own
+            if user_id_profile and user.id != user_id_profile and not user.allowed(Role.ADMIN):
```
anthonygego:
```suggestion
```

decorators.py
```diff
@@ -16,16 +16,24 @@ def decorated_function(*args, **kwargs):
     return decorated_function
 
 
-def check_access_level(access_level):
+def check_access_level(*allowed_roles):
     def decorator(f):
         @wraps(f)
         def decorated_function(*args, **kwargs):
+            user_id_profile = kwargs.get('user_id')
```
anthonygego:
```suggestion
```

You'd better do the check in the appropriate route processing function so that `user_id` has the right meaning. Here you will only check for roles.

user.py
```diff
@@ -113,6 +114,7 @@ def user_profile(user_id, current_year):
 
 @user_bp.route('/update_user_profile', methods=['POST'])
 @login_required
+@check_access_level(Role.ADMIN, Role.RESEARCHER, Role.TEACHER)
 def update_user_profile():
```
anthonygego: ...and therefore put the id check below.

anthonygego (changes requested):

anthonygego:
Can you rebase ?

user.py
```diff
@@ -113,11 +123,17 @@ def user_profile(user_id, current_year):
...
     form = request.form
     if not form:
         return make_response("Problem with form request", 500)
 
+    user_id = request.form['user_id']
```
anthonygego: Beware, you fetch a string from the form and your `is_allowed_user` expects an int. I recommend you to set the user id in the route. This way, you don't need a hidden field in your form and just need to update the `action` attribute.

user.py
```diff
@@ -88,9 +88,19 @@ def users(user_type):
     return render_template('users.html', users=all_users, user_type=user_type)
 
 
+def is_allowed_user(user_id):
+    user = db.session.query(User).filter_by(id=session["user_id"]).first()
+    return user.id != user_id and not user.allowed(Role.ADMIN)
```
anthonygego: `user.id` should always be the same as `session["user_id"]`. You can compare them directly.

anthonygego (changes requested):

user.py
```diff
@@ -90,9 +90,19 @@ def users(user_type):
     return render_template('users.html', users=all_users, user_type=user_type)
 
 
+def is_allowed_user(user_id):
+    user = db.session.query(User).filter_by(id=session["user_id"]).first()
+    return user.id != session["user_id"] and not user.allowed(Role.ADMIN)
```
```suggestion
    return user_id == session["user_id"] or session["is_admin"]
```

The function name indicates the opposite of what it is meant to do. Your first condition is however always false here. You can actually rely on session information only.

user.py
```diff
@@ -90,9 +90,19 @@ def users(user_type):
...
+
+
 @user_bp.route('/profile/<int:user_id>/<int:current_year>')
 @login_required
+@check_access_level(Role.ADMIN, Role.RESEARCHER, Role.TEACHER)
 def user_profile(user_id, current_year):
+    if is_allowed_user(user_id):
```
```suggestion
    if not is_allowed_user(user_id):
```

user.py
```diff
@@ -113,13 +123,18 @@ def user_profile(user_id, current_year):
...
+@user_bp.route('/update_user_profile/<int:user_id>', methods=['POST'])
 @login_required
-def update_user_profile():
+@check_access_level(Role.ADMIN, Role.RESEARCHER, Role.TEACHER)
+def update_user_profile(user_id):
     form = request.form
     if not form:
         return make_response("Problem with form request", 500)
 
+    if is_allowed_user(user_id):
```
```suggestion
    if not is_allowed_user(user_id):
```

anthonygego (changes requested):

PR #22, opened 2024-07-30
Issues#19 Internal server error when adding existing user account.

SamuelVch98:
Add a static method to the database to check whether the email already exists and, if so, display an error.


db.py
```diff
@@ -48,6 +51,16 @@ def allowed(self, access_level):
         else:
             return False
 
+    @staticmethod
+    def validate_unique_email(mapper, connect, target):
+        if target.email:
+            query = db.session.query(User).filter(User.email == target.email, User.id != target.id).exists()
+            if db.session.query(query).scalar():
```
```suggestion
            if db.session.query(User).filter(User.email == target.email).first()
```
`target.id` is not set yet at this stage and is equals to `None`. The id is set after insertion as it is auto-incremental. By using `first()`, you will get the user or `None` if it does not exists, this is more direct.

db.py
```diff
@@ -48,6 +51,16 @@ def allowed(self, access_level):
...
+                raise ValueError("Email already exists")
+
+
+event.listen(User, 'before_insert', User.validate_unique_email)
```
anthonygego: Defining a hook is useful in case you don't exactly know where the insertions will be done, or in case there are several known locations.

In the context of this app user insertion is done at two locations : when the user has logged in by themselve, or when they are added by the administrators. In the first case, the check is performed to know if the account must be created. So there is only one location you may have a conflict.

Therefore, maybe it is more straightforward and **explicit** to perform this check in the appropriate route.

According to the User model definition, the only integrity error that may arise is a duplicate email. You can also fully rely on exception handling and return a comprehensive message to the user in case an IntegrityError has been raised. This is not defensive but not wrong neither, as it is just the propagated result of the DBMS inner check.

user.py
```diff
@@ -64,6 +65,9 @@ def add_user():
                                         max_loads=max_load)
             db.session.add(new_researcher)
             db.session.commit()
+    except ValueError as e:
```
anthonygego: So here either you make the check before adding the new user to the `db.session`, either you catch an `sqlachemy.exc.IntegrityError` and return that the user already exists in database as your flash message.

db.py
```diff
@@ -3,6 +3,8 @@
 from flask import Flask
 from flask_sqlalchemy import SQLAlchemy
 import json
+
+from sqlalchemy import event
```
```suggestion
```

PR #25, opened 2024-09-18
Issue#23 a user can easily change their status in their profile

SamuelVch98:
prevent a non-admin from changing his status by adding a condition that checks that the user is an admin


user.py
```diff
@@ -167,16 +167,17 @@ def update_user_profile(user_id):
...
-        user.is_researcher = is_researcher
-        user.supervisor_id = supervisor_id
-        if is_researcher:
-            if researcher is None:
-                new_researcher = Researcher(user_id=user.id, researcher_type=researcher_type, max_loads=max_loads)
-                db.session.add(new_researcher)
-            else:
-                researcher.max_loads = max_loads
-                researcher.researcher_type = researcher_type
+        if user.admin:
```
anthonygego, line 170: Here you check the user that is modified is an admin. You should check that the session user is an admin.

PR #32, opened 2024-09-25
Issue#24 current year coherency

SamuelVch98:
The current year was defined in two different ways and this posed a problem of consistency. I defined the current year statically using the ‘Configuration’ table.


util.py
```diff
@@ -1,10 +1,14 @@
...
-    current_date = datetime.date.today()
-    current_year = current_date.year
-    if current_date.month >= 9:  # Assuming academic year starts in September
-        return current_year + 1
-    else:
-        return current_year
+    try:
+        current_year = Configuration.query.filter_by(is_current_year=True).first()
+        return current_year.year
+    except Exception as e:
```
anthonygego: There is no need to catch an exception if you don't perform any handling. Just leave the code as is, and do exception handling at the higher level.

templates/layout.html
```diff
@@ -99,7 +99,14 @@
...
+                                                data-message="Are you sure you want to add a new year?"
+                                                data-url="{{ url_for('config.new_year') }}">Add New Year
+                                        </button>
+                                        <button class="dropdown-item" data-bs-toggle="modal"
+                                                data-bs-target="#confirmationModal"
+                                                data-message="Are you sure you want to move on to the next year?"
+                                                data-url="{{ url_for('config.next_year') }}">Move on to the next year
                                         </button>
                                     </li>
                                 </ul>
```
anthonygego: It could be nice to have the current academic year shown somewhere.

config.py
```diff
@@ -22,10 +22,31 @@ def next_year():
...
+@config_bp.route('/next_year', methods=['POST'])
+@login_required
+@check_access_level(Role.ADMIN)
+def next_year():
+    current_year = Configuration.query.filter_by(is_current_year=True).first()
+    try:
+        is_new_year = Configuration.query.filter_by(year=current_year.year + 1).first()
+        # Check if the following year already exists
+        if is_new_year is None:
+            return make_response("Please create a new year before moving on to the next one", 500)
```
anthonygego: The entry can probably be inserted automatically if needed to improve user friendlyness. Besides, `make_response` is not a very elegant way to return feedback to the user.

config.py
```diff
@@ -22,10 +22,31 @@ def next_year():
...
         db.session.rollback()
         raise e
 
     return redirect(url_for("index"))
+
+# This method allows the admin to move on to the next academic year for all users.
+@config_bp.route('/next_year', methods=['POST'])
+@login_required
+@check_access_level(Role.ADMIN)
+def next_year():
```
anthonygego: @luciledierckx Using this implementation we'll always be going in one way. Is there any case for which we'd like to revert and go to the past ?

course.py
```diff
@@ -135,8 +135,7 @@ def search_teachers():
 @login_required
 @check_access_level(Role.ADMIN)
 def course_info(course_id):
-    dynamic_year = get_current_year()
-    current_year = int(request.args.get('current_year')) if request.args.get('current_year') else dynamic_year
+    current_year = int(request.args.get('current_year'))
```
anthonygego: Please avoid putting required arguments in request args: put them in the route, there are automatically sanitized.

config.py
```diff
@@ -22,10 +22,31 @@ def next_year():
...
+        if is_new_year is None:
+            return make_response("Please create a new year before moving on to the next one", 500)
+
+        current_year.is_current_year = False
+        is_new_year.is_current_year = True
+        db.session.commit()
+
+    except Exception as e:
+        db.session.rollback()
+        raise e
```
anthonygego: Maybe you can return a feedback to the user in this case.

anthonygego (changes requested):

config.py
luciledierckx: The only reason we would like to revert would be that we mistakenly call the next_year functionality more than once a year. As this action is available for all admins, maybe it would be safer to have an option to go back in case one of them mistakenly calls it?

config.py
```diff
@@ -6,26 +6,59 @@
...
 
         config = Configuration(year=new_year)
         db.session.add(config)
         db.session.commit()
-        Configuration.update_current_year(config.id)
 
     except Exception as e:
         db.session.rollback()
-        raise e
+        flash("An error occurred while processing your request. Please try again.", "error")
```
anthonygego: Maybe we can log the error or at least give an insight in the feedback of what's gone wrong. There is almost no chance that "trying again" will work.

config.py
```diff
@@ -6,26 +6,59 @@
...
+        else:
+            # If the new year exists, just update the current year status
+            is_new_year.is_current_year = True
+
+        current_year.is_current_year = False
+        db.session.commit()
+
+    except Exception as e:
+        db.session.rollback()
+        flash("An error occurred while processing your request. Please try again.", "error")
```
anthonygego: Same remark here.

config.py
```diff
@@ -6,26 +6,59 @@
...
+
+
+# This method allows the admin to move on to the next academic year for all users.
+@config_bp.route('/next_year', methods=['POST'])
+@login_required
+@check_access_level(Role.ADMIN)
+def next_year():
+    current_year = Configuration.query.filter_by(is_current_year=True).first()
+    try:
+        is_new_year = Configuration.query.filter_by(year=current_year.year + 1).first()
```
anthonygego: `is_new_year` => `new_year`. It is confusing as it is not a boolean.

templates/layout.html
```diff
@@ -99,17 +99,33 @@
...
-                        <a href="{{ url_for('auth.logout') }}" class="btn btn-outline-danger ms-2">Logout</a>
+                    <div class="d-flex align-items-center ms-auto">
+                        <div class="me-0">
+                            <div class="navbar-text me-2">
+                                <input type="text" class="form-control form-control-sm" id="dynamicYearInput"
+                                       value="{{ dynamic_year }} - {{ dynamic_year + 1 }}" style="width: 100px;" disabled>
+                            </div>
+                        </div>
+
+                        <div class="navbar-text ms-0"> <!-- Ajustez ici -->
```
```suggestion
                        <div class="navbar-text ms-0">
```

config.py
```diff
@@ -2,30 +2,97 @@
...
+        db.session.rollback()
+        logging.error(f'An error occurred: {str(e)}', exc_einfo=True)
+        flash("An error has occurred. We will solve it as soon as possible", "error")
+
+        return redirect(url_for("config.manage_years"))
+
+    return redirect(url_for("config.manage_years"))
+
+
+@config_bp.route('/delete_year/<int:id_year>', methods=['POST'])
```
AlexandreDoneux: You can't delete a year if you already attributed it as being the current year. That doesn't solve the problem mentioned by Lucile. 

Moreover, we don't really want to delete a year, we just want to change the current year to a previous year. You **could** adapt the `/next_year` endpoint into a `/change_year` where you can specify a certain year. That way, you don't need a second endpoint to go to a previous year.

config.py
```diff
@@ -2,30 +2,97 @@
...
+            # If the new year exists, just update the current year status
+            new_year.is_current_year = True
+
+        current_year.is_current_year = False
+        db.session.commit()
+
+    except Exception as e:
+        db.session.rollback()
+        logging.error(f'An error occurred: {str(e)}', exc_einfo=True)
+        flash("An error has occurred. We will solve it as soon as possible", "error")
```
AlexandreDoneux: Not a helpful error message.

config.py
```diff
@@ -2,30 +2,97 @@
...
+        if config.is_current_year:
+            flash("You cannot delete the current year", "error")
+        else:
+            db.session.delete(config)
+            db.session.commit()
+            flash("Year deleted successfully", "success")
+    except Exception as e:
+        db.session.rollback()
+        logging.error(f'An error occurred while deleting year {id_year}: {str(e)}', exc_info=True)
+        flash("An error occurred. Please try again.", "error")
```
AlexandreDoneux: Not a helpful message.

config.py
```diff
@@ -2,30 +2,97 @@
...
+
+        current_year.is_current_year = False
+        db.session.commit()
+
+    except Exception as e:
+        db.session.rollback()
+        logging.error(f'An error occurred: {str(e)}', exc_einfo=True)
+        flash("An error has occurred. We will solve it as soon as possible", "error")
+
+        return redirect(url_for("config.manage_years"))
```
AlexandreDoneux: You don't need that redirect as it is already present after your try-except

AlexandreDoneux (changes requested):
Change some error messages and change the way you handle year "deletion"

AlexandreDoneux:
Looks good to me, waiting for @anthonygego approval

anthonygego (approved):

PR #39, opened 2024-10-01
Issue#33 newly registered users can't edit their profile

SamuelVch98:
The application did not have a role for classic users. I added a new role to allow all users to have access to their profile.


user.py
```diff
@@ -100,7 +100,7 @@ def is_allowed_user(user_id):
 
 @user_bp.route('/profile/<int:user_id>/<int:current_year>')
 @login_required
-@check_access_level(Role.ADMIN, Role.RESEARCHER, Role.TEACHER)
+@check_access_level(Role.USER)
```
anthonygego: There is no need to define such a role. You can simply allow any logged user and remove this decorator.

db.py
```diff
@@ -40,14 +41,14 @@ def validate_active(self, key, value):
...
-        if access_level == Role.ADMIN:
-            return self.admin
-        elif access_level == Role.RESEARCHER:
-            return self.is_researcher
-        elif access_level == Role.TEACHER:
-            return self.is_teacher
-        else:
-            return False
+        role_access = {
+            Role.USER: True,
```
anthonygego: You can therefore remove this one and let `False` be the default value. Changing this to a dictionary lookup remains a good idea.

anthonygego (changes requested):

db.py
```diff
@@ -12,6 +12,7 @@ class Role(Enum):
     ADMIN = "admin"
     RESEARCHER = "researcher"
     TEACHER = "teacher"
+    USER = "user"
```
AlexandreDoneux: Remove this line, not needed anymore

db.py
```diff
@@ -40,14 +41,13 @@ def validate_active(self, key, value):
...
-        if access_level == Role.ADMIN:
-            return self.admin
-        elif access_level == Role.RESEARCHER:
-            return self.is_researcher
-        elif access_level == Role.TEACHER:
-            return self.is_teacher
-        else:
-            return False
+        role_access = {
+            Role.ADMIN: self.admin,
```
AlexandreDoneux: Keep consistency in your variable names : use `self.is_admin`

db.py
```suggestion

```

user.py
```diff
@@ -105,13 +105,13 @@ def is_allowed_user(user_id):
...
```
```suggestion
```

anthonygego:
https://github.com/UCL-INGI/ictm-teaching/blob/22afecd73629950459edeadc240ca6b4c2c87bcf/user.py#L134-L136
also need to be adapted.

anthonygego (approved):

PR #42, opened 2024-10-03
Issue#34 researcher preferences should be put apart

SamuelVch98:
I've removed the ‘preferences’ tab from the profile and incorporated it into a new page. I only display the preferences for this year in the profile


anthonygego:
The diff seems to include code from #32. I'll wait for the review.

anthonygego:
Can you solve conflicts ?

templates/layout.html
```diff
@@ -92,8 +98,8 @@
                                 </ul>
                             </li>
                             <li>
-                            <a class="nav-link"
-                               href="{{ url_for('config.manage_years', configurations=configurations) }}">Configurations</a>
+                                <a class="nav-link"
+                                   href="{{ url_for('config.manage_years', configurations=configurations) }}">Configurations</a>
```
```suggestion
                                   href="{{ url_for('config.manage_years') }}">Configurations</a>
```

course_preference.py
```diff
@@ -34,29 +33,30 @@ def save_preference():
 
     preferences = data['preferences']
```
anthonygego:
```suggestion
    preferences = data.get('preferences', None)
```

You should also consider the case the key is not defined.

course_preference.py
```diff
@@ -34,29 +33,30 @@ def save_preference():
...
     researcher = Researcher.query.filter(Researcher.user_id == user_id).first()
     if researcher is None:
-        return make_response("User is not a researcher", 500)
+        flash("Researcher not found", "error")
 
-    new_course_ids = {preference['course_id'] for preference in preferences}
-    delete_old_preferences(researcher.id, new_course_ids, current_year)
+    delete_old_preferences(researcher.id, current_year)
 
+    rank = 0
```
anthonygego:
```suggestion
```

course_preference.py
```diff
@@ -34,29 +33,30 @@ def save_preference():
...
     for preference in preferences:
         try:
+            rank += 1
```
anthonygego:
```suggestion
```

course_preference.py
```diff
@@ -34,29 +33,30 @@ def save_preference():
...
     if researcher is None:
-        return make_response("User is not a researcher", 500)
+        flash("Researcher not found", "error")
 
-    new_course_ids = {preference['course_id'] for preference in preferences}
-    delete_old_preferences(researcher.id, new_course_ids, current_year)
+    delete_old_preferences(researcher.id, current_year)
 
+    rank = 0
     for preference in preferences:
```
anthonygego:
```suggestion
    for rank, preference in enumerate(preferences):
```

user.py
```diff
@@ -137,8 +140,8 @@ def user_profile(user_id, current_year):
     preferences = []
     if researcher:
         preferences = db.session.query(PreferenceAssignment).filter_by(researcher_id=researcher.id,
-                                                                       course_year=current_year).all()
-
+                                                                       course_year=current_year).order_by(
+            PreferenceAssignment.rank).all()
```
anthonygego: What is the title of this PR ?

user.py
```diff
@@ -157,6 +160,32 @@ def delete_researcher(user_id):
         db.session.commit()
 
 
+@user_bp.route('/<int:user_id>/preferences/<int:current_year>', methods=['GET'])
+@login_required
+@check_access_level(Role.ADMIN, Role.RESEARCHER)
+def preferences(user_id, current_year):
+    if not is_allowed_user(user_id):
```
anthonygego: You should also check the user is a researcher. The page content has no sense otherwise, and moreover, the template rendering fails as `researcher` is None.

templates/user_profile.html
```diff
@@ -150,140 +146,33 @@ <h2>Your informations</h2>
...
-                                {% endfor %}
-                                </tbody>
-                            </table>
-                        {% endif %}
-                    </div>
-                </div>
-            {% endif %}
         </div>
+        <br>
+        {% if researcher %}
```
anthonygego: What is the title of this PR ?

anthonygego (changes requested):

templates/preferences.html
```diff
@@ -0,0 +1,210 @@
...
+                        {% if not session.is_admin %}
+                            <td>
+                                <a href="{{ url_for('course_preference.delete_preference', preference=preference.id) }}"
+                                   class="delete-icon">
+                                    <i class="fas fa-trash delete-icon"></i>
+                                </a>
+                            </td>
+                        {% else %}
+                            <td></td>
+                        {% endif %}
```
anthonygego:
```suggestion
                        <td>
                            {% if not session.is_admin %}
                                <a href="{{ url_for('course_preference.delete_preference', preference=preference.id) }}"
                                   class="delete-icon">
                                    <i class="fas fa-trash delete-icon"></i>
                                </a>
                            {% endif %}
                        </td>
```

templates/preferences.html
```diff
@@ -0,0 +1,210 @@
...
+                    </tr>
+                {% endfor %}
+                </tbody>
+            </table>
+
+            <table class="table table-hover" id="editablePreferencesTable" style="display: none;">
+                <thead>
+                <tr>
+                    <th>Code</th>
```
anthonygego:
```suggestion
                    <th colspan="2">Code</th>
```

course_preference.py
```diff
@@ -29,34 +29,32 @@ def save_preference():
     current_year = get_current_year()
 
     if data is None:
-        return make_response("No data received", 500)
```
anthonygego, line 32: I don't remember the discussions... you call your `save_preference` route using AJAX. Using `flash` will produce an HTTP 200 code as the page will be rendered. You will therefore never show the error to the user.

Either you can get back to error codes or produce json. Then, catch the error message in your ajax call to show it to the user.

templates/preferences.html
```diff
@@ -0,0 +1,210 @@
...
+                    type: 'POST',
+                    url: "{{ url_for('course_preference.save_preference') }}",
+                    contentType: 'application/json',
+                    data: JSON.stringify(data),
+                    success: function (response) {
+                        alert('Preferences updated successfully.');
+                        location.reload(); // Refresh the page to show the updated preferences
+                    },
+                    error: function (error) {
+                        alert('Error updating the preferences.');
```
anthonygego: Here an error message will be shown only for the unmanaged errors in the python code, that is, when the app crashes.

templates/preferences.html
```diff
@@ -0,0 +1,210 @@
...
+            });
+
+            $('#savePreferencesButton').click(function () {
+                const preferences = [];
+                !$('#editablePreferencesTable tbody tr').not('#template-row').each(function () {
+                    const courseValue = $(this).find('select.course-select').val();
+
+                    if (courseValue) {
+                        const [course_id, course_year] = courseValue.split("-");
+                        preferences.push({course_id: course_id, course_year: course_year});
```
anthonygego: Do we really need to pass the year as argument for each course ? Preferences are registered for one year at the time.

It seems we could end up, by modifying the form content, to edit past and future preferences, if the course is defined. Moreover, the `save_preference` route seems to rely on the `current_year` for part of its operations.

anthonygego (changes requested):

templates/preferences.html
```diff
@@ -0,0 +1,211 @@
...
+                    <th colspan="2">Code</th>
+                </tr>
+                </thead>
+                <tbody class="table-group-divider">
+                {% for preference in preferences %}
+                    <tr>
+                        <td>
+                            <select class="form-control course-select">
+                                {% for course in courses %}
+                                    <option value="{{ course.id }}-{{ course.year }}"
```
anthonygego: This should be updated too...

anthonygego (changes requested):

anthonygego (approved):

PR #45, opened 2024-10-09
Fix assign teachers

SamuelVch98:
When I assigned two teachers to a course and then deleted one of them, the deletion didn't happen.


anthonygego:
The diff seems to include code from #32. I'll wait for the review.

anthonygego:
Can you solve conflicts ?

course.py
```diff
@@ -125,10 +125,8 @@ def search_teachers():
     if not validate_string_pattern(search_term):
         return make_response("Invalid search term", 400)
 
-    teachers = db.session.query(User).join(Teacher).filter(
-        User.active == True,
-        User.name.ilike(f'%{search_term}%')
-    ).all()
+    teachers = db.session.query(User).filter(User.active == True, User.is_teacher == True,
+                                             User.name.ilike(f'%{search_term}%')).all()
```
anthonygego: I think it could be useful, either to : 
- specify the search is on the last name only.
- search in both name and firstname, as the complete name is chown.
- apply the `LIKE` operator on the concatenated full name
  `SELECT * FROM user WHERE CONCAT(name, ' ', first_name) LIKE '%name firs%';`

anthonygego (approved):

PR #46, opened 2024-10-10
Issue#27 researcher profile always kept in databae

SamuelVch98:
Suppression of the status of researcher and professor in the DB


anthonygego:
Can you solve conflicts ? Thanks

anthonygego (commented):
I don't exactly know what to think here. It seems you keep a (NULL,NULL) entry in database to know if a user is a teacher.

However, we don't need that. Someone who has admin rights on a course is a teacher and will be allowed to see their course information.

@luciledierckx do we need to keep a track of who is a teacher or not in case no course is assigned to them ?

Besides, there actually is no check for course information access. Only ADMIN role can show the page. In practice, teachers should be able to show their (and only their) course information.

SamuelVch98:
The only purpose of this is to be able to view a teacher in the list of teachers who has not yet been assigned a lesson. This is what I was advised to do

anthonygego:
> The only purpose of this is to be able to view a teacher in the list of teachers who has not yet been assigned a lesson. This is what I was advised to do

If this is the case therefore  the `is_teacher` column in the `User` model is relevant and can be kept.

luciledierckx:
> > The only purpose of this is to be able to view a teacher in the list of teachers who has not yet been assigned a lesson. This is what I was advised to do
> 
> If this is the case therefore the `is_teacher` column in the `User` model is relevant and can be kept.

This seems good to me too!

db.py
```diff
@@ -42,7 +41,7 @@ def validate_active(self, key, value):
     def allowed(self, access_level):
         role_access = {
             Role.ADMIN: self.is_admin,
-            Role.RESEARCHER: self.is_researcher,
+            Role.RESEARCHER: self.user_researcher is not None,
```
anthonygego: You can probably rename the backref to `researcher_profile` so that it is more intuitive to use `User.researcher_profile`

templates/user_profile.html
```diff
@@ -105,7 +105,7 @@ <h2>Your informations</h2>
                                     <div class="form-check">
                                         <input type="checkbox" class="form-check-input" id="is_researcher"
                                                name="is_researcher"
-                                               {% if requested_user.is_researcher %}checked{% endif %}
+                                               {% if is_researcher or researcher %}checked{% endif %}
```
anthonygego: Why is this double condition necessary ? It seems that if `researcher` is defined, `is_researcher` will always be `True`.

user.py
```diff
@@ -9,12 +9,32 @@
 user_bp = Blueprint('user', __name__)
 
 
+def create_teacher(user_id):
+    if db.session.query(Teacher).filter_by(user_id=user_id).first() is None:
+        new_teacher = Teacher(user_id=user_id)
+        db.session.add(new_teacher)
+        db.session.commit()
```
```suggestion
```
Do you still need this one ?

templates/user_profile.html
SamuelVch98: Yes, that's right, is_researcher was based on the logged-in user, whereas researcher was based on the user whose profile you wanted to view.

user.py
```diff
@@ -130,6 +142,20 @@ def user_profile(user_id, current_year):
...
+
+
+def delete_teacher(user_id):
+    teacher = db.session.query(Teacher).filter(Teacher.user_id == user_id).first()
+    if teacher:
+        db.session.delete(teacher)
+        db.session.commit()
+
+
```
```suggestion
```

PR #47, opened 2024-10-16
Feature publish assignment

SamuelVch98:
Assignments can be published for everyone or just for teachers
Display assignments in the researcher/teacher view


anthonygego:
Can you rebase ?

db.py
```diff
@@ -133,6 +148,11 @@ class Year(db.Model):
     id = db.Column(db.Integer, primary_key=True)
     year = db.Column(db.Integer, nullable=False)
     is_current_year = db.Column(db.Boolean, default=False)
+    publication_status = db.Column(
+        Enum(*PUBLICATION_STATUS, name='publication_status_enum'),
+        nullable=False,
+        default='Draft'
+    )
```
AlexandreDoneux: I don't understand why have a publication status inside the year table. 

The assignements we consider up to date are the ones in the `assignement_published` table. When a draft is published all the assignements that were in the draft tables should be copied to the published table. So what is the use of `publication_status` ?

AlexandreDoneux: I had a talk with Anthony. We aim to only keep the two assignment tables and remove the `publication_status` from the Year table. That way one can make changes on a published assignment and save it as draft, later publishing it.

AlexandreDoneux: We just need to handle the case where we want the assignment to be visible only for teachers.

app.py
```diff
@@ -53,11 +55,40 @@ def inject_configurations():
...
-    courses_teacher = db.session.query(Course).filter_by(year=current_year).join(Teacher).filter(Teacher.user_id == user.id).all()
-    return render_template("home.html", user=user, courses=courses_teacher, evaluations=user.evaluations)
+    researcher = db.session.query(Researcher).filter_by(user_id=user.id).first()
+    teacher = db.session.query(Teacher).filter_by(user_id=user.id).first() if user.is_teacher else None
+
+    # Get the teacher's courses
+    courses_teacher = db.session.query(Course).join(Teacher).filter(
+        and_(Teacher.user_id == user.id, Course.year == current_year)
+    ).all() if year.publication_status != 'Draft' and user.is_teacher else []
+
```
AlexandreDoneux: Why check if the user is a teacher ? If they are not your join between the Teacher and Course tables should be empty. So there is no need to make that check.

And why do a check if the publication status from the current_year is not a draft. As mentioned in another comment, having the publication_status in the Year table is irrelevant. Are you trying to filter to only display the courses that have been assigned to researchers ? I don't think you need it, but if you really want that you need to check it through the `assignement_published` table.

Because for the moment, courses that have not been assigned to researchers can't be viewed by the teachers.

app.py
```diff
@@ -53,11 +55,40 @@ def inject_configurations():
...
 @login_required
-def index():  # put application's code here
+def index():
     current_year = get_current_year()
+    year = db.session.query(Year).filter_by(year=current_year).first()
     user = db.session.query(User).filter_by(email=session['email']).first()
-    courses_teacher = db.session.query(Course).filter_by(year=current_year).join(Teacher).filter(Teacher.user_id == user.id).all()
-    return render_template("home.html", user=user, courses=courses_teacher, evaluations=user.evaluations)
+    researcher = db.session.query(Researcher).filter_by(user_id=user.id).first()
+    teacher = db.session.query(Teacher).filter_by(user_id=user.id).first() if user.is_teacher else None
```
AlexandreDoneux: I feel like you could use conditionals on the user role to avoid unnecessary fetching from the DB and computing.

ex : 
```py
if user.is_teacher : 
    # fetch and computing related to teacher
elif user.is_researcher : 
    # fetch and computing related to researcher
elif ... : 
   # ...
```

AlexandreDoneux (changes requested):
Changes are needed regarding the way draft and published assignments work.

Also use conditional fetching from DB and computing.

db.py
SamuelVch98: How do we manage the case of publication for teachers only?

app.py
```diff
@@ -53,11 +55,49 @@ def inject_configurations():
...
```
AlexandreDoneux: You can also handle these with conditionals. If the user is a teacher no need to check if they are a researcher. That way we have one less call to the database.

app.py
```diff
@@ -53,11 +55,49 @@ def inject_configurations():
...
+    # Teacher data
+    courses_teacher = []
+    researcher_supervised = []
+
+    # Researcher data
+    researcher_courses = []
+    researcher_current_course = []
+    researcher_evaluations = []
+
+    if teacher:
```
AlexandreDoneux:
```suggestion
    if user.is_teacher :
       ...
    else : 
       researcher = db.session.query(Researcher).filter_by(user_id=user.id).first()
       ...
```

No need to go fetch the `teacher` in the database because you don't use it later. And you can check if the user is a researcher here in the condition as suggested in the comment above.

app.py
```diff
@@ -53,11 +55,49 @@ def inject_configurations():
...
+        # Get the courses where the researcher is assigned
+        researcher_courses = researcher.assigned_courses
+        researcher_current_course = [
+            course for course in researcher_courses if course.year == current_year
+        ]
+        researcher_evaluations = researcher.user.evaluations
+
+    return render_template("home.html", user=user, courses_teacher=courses_teacher,
+                           researcher_supervised=researcher_supervised, researcher_courses=researcher_courses,
+                           researcher_current_course=researcher_current_course, evaluations=researcher_evaluations)
```
AlexandreDoneux: I think it is best to group all those variables (apart from user maybe) in one dict named `data` for example. That way you improve readability and if another variable is needed you don't need to add it in the render_template.

I would also change some names (`courses_teacher` to `teacher_courses` and `researcher_supervised` to `supervised_researchers` (for proper English), `researcher_current_courses` (missing "s"), `researcher_evaluations`).

AlexandreDoneux: Maybe it is a bit of a stretch but in that  `data` dict we can store these data only if they are needed. For example if the user is a teacher we don't need the data related to researchers. And because you are using conditional blocks in the templates it shouldn't be a problem.

templates/home.html
```diff
@@ -8,66 +9,78 @@
...
-        <div class="container-fluid">
-            <div class="card">
+
+    <div class="container-fluid">
+        <!-- Header Section -->
+        <h1 class="mt-3 mb-3">{{ session.first_name }} {{ session.name }}</h1>
+        <hr>
+
+        <!-- Researcher Section -->
+        {% if is_researcher %}
```
AlexandreDoneux, line 22: Maybe do role checks in a more harmonized way. You are checking if the user is an admin through session, if they are a teacher through a User result from the DB and if they are a researcher like this. How about storing the role in session @anthonygego ?

Also, where is `is_researcher` defined ?

AlexandreDoneux (changes requested):

app.py
SamuelVch98: I'm going to have to add an if condition in the else if the user is neither a teacher nor a researcher

app.py
SamuelVch98: We hadn't discussed the fact that it was possible to be a teacher and a researcher at the same time?

templates/home.html
SamuelVch98: Its defined in the context_processor

app.py
AlexandreDoneux: I'll ask @anthony. But even if that is the case, we better do these calls in conditions. We could use the session to check the roles the user has.

templates/home.html
AlexandreDoneux: Thanks, I will have a look

app.py
AlexandreDoneux: That is a possibility. You can use an if/elif/else checking on the session flags for example

app.py
SamuelVch98: If it's OK with @anthonygego, I can store this in the session

SamuelVch98: We still have a problem differentiating between a global assignment and an assignment for teachers only.

PR #48, opened 2024-10-21
Feature colour preferences

SamuelVch98:
Calculate the number of times a user has given a lesson to colour their preferences


anthonygego:
Can you rebase ?

